PR #24, opened 2015-07-27
Bibliography fixes

rhattersley:

PR #26, opened 2015-07-27
Front matter fixes

rhattersley:
NB. The bibliography references are dependent on #24.


marqh:
#24 has merge conflicts, so i'm waiting on those to be resolved before merging that one, then this one to follow

marqh:
ok, #24 is in so I'm assuming this is good to go, it seems to be fine to merge

PR #28, opened 2015-10-26
Chapter 2 fixes

davidhassell:
Hi Richard and Mark,

Here are some minor format suggestions for chapter 2: spaces before commas and the like. CF-1.5 changed to CF-1.6, also.

Thanks and all the best,

David


rhattersley:
Thanks @davidhassell :smile:

PR #29, opened 2015-10-26
Chapter3 fixes

davidhassell:
Hi Richard and Mark,

More minor typesettting ideas  for chapter 3.

All the best,

David


ch03.adoc
```diff
@@ -295,7 +295,7 @@ The attributes **`flag_values`**,
 
 The **`flag_values`** and **`flag_meanings`**
       attributes describe a status flag consisting of mutually exclusive coded values. The **`flag_values`** attribute is the same type as the variable to which it is attached, and contains a list of the possible flag values. The **`flag_meanings`** attribute is a string whose value is a blank separated list of descriptive words or phrases, one for each flag value. Each word or phrase should consist of characters from the alphanumeric
-      set and the following five: '_', '-', '.', '+', '@'. If multi-word phrases are used to describe the flag values, then the words within a phrase should be connected with underscores. The following example illustrates
+      set and the following five: "_","-", ".","+", "@". If multi-word phrases are used to describe the flag values, then the words within a phrase should be connected with underscores. The following example illustrates
```
rhattersley: I don't think this line needs changing.

1. Both the [HTML](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#flags) and [PDF](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.pdf) versions use a single quote character.
2. You've lost a couple of the spaces between items.

rhattersley:
> Chapter 2 fixes

You've accidentally included the changes from #28.

davidhassell:
Hi Richard,

No problem! The double quotes ("+", "-", ...) were a compromise: The existing 1.6 has single quotes as you say, but the single quotes in asciidoc give emphasised text with no quotes. Some sort quotes are needed (so the commas are not in the list!) - can quotes be escaped?

David

davidhassell:
Re. including previous chapter's changes - oops, sorry. I probably did that with chapters 4 and 5, too. Won't happen again (touch wood).

David

rhattersley:
To get rid of the extraneous commit you need to rebase. Assuming you have `upstream` pointing at this repo, `origin` pointing at your fork, and a clean checkout in your local repo:
```
$ git checkout chapter3_fixes
$ git fetch upstream
$ git rebase upstream/master
$ git push origin +chapter3_fixes
```

ch03.adoc
rhattersley: > single quotes in asciidoc give emphasised text with no quotes

As of asciidoctor 1.5 (as used by the automated build) this is [no longer true](https://github.com/asciidoctor/asciidoctor/issues/717). You can see this at http://cf-metadata.github.io/#flags. So the simplest thing to do here is nothing.

If you're stuck with an earlier version then perhaps the back-tick - forward-tick technique (as described for ["Single quoted text"](http://www.methods.co.nz/asciidoc/chunked/ch10.html)) would allow it to work in both versions.

PR #33, opened 2015-10-27
chapter 6 fix

davidhassell:


ch06.adoc
```diff
@@ -30,7 +30,7 @@ string-valued auxiliary coordinate variables. The **`coordinates`** attribute
       other dimensions are axis dimensions. If a character variable has only
       one dimension (the maximum length of the string), it is regarded as a
       string-valued scalar coordinate variable, analogous to a numeric scalar
-      coordinate variable (see <<scalar-coordinate-variables>>)
+      coordinate variable (see <<scalar-coordinate-variables>>).
```
rhattersley: The missing full-stop is also missing in the original HTML and PDF versions. Perhaps this fix would be better targetted at v1.7?

rhattersley:
Hi @davidhassell. Now that this repo has been officially blessed I guess you should really put this change through [trac](http://cf-trac.llnl.gov/trac) ... that seems *ridiculously* long-winded though! :scream:

@JonathanGregory - how about we cut out the red tape for such a simple "bug fix"?

JonathanGregory:
I agree. If we had a ticket it would be a defect ticket and we've never had one for fixing just punctuation. This is not a change of content, just editorial, and I'm sure it can be done with no more discussion.

rhattersley:
@davidhassell - please could you rebase.

> This is not a change of content, just editorial, and I'm sure it can be done with no more discussion.

@JonathanGregory - I'm thinking this doesn't even warrant an entry in the Revision History, agreed?

JonathanGregory:
Yes, I agree. Thank you and Mark for implementing tickets. Maybe there will be one left for me to try!

rhattersley:
> Maybe there will be one left for me to try!

Oh, there are still plenty! :wink:

rhattersley:
Thanks @davidhassell :+1:

PR #41, opened 2015-11-22
Manual table captions

### LINKED ISSUE #6: Per-section table numbering
See cf-metadata/convert#14

rhattersley:
Closes #6.

### DIFF
--- a/appa.adoc
+++ b/appa.adoc
@@ -6,8 +6,10 @@
 
 All CF attributes are listed here except for those that are used to describe grid mappings. See Appendix F for the grid mapping attributes.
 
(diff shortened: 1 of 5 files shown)


marqh:
good stuff, i get all the correct appendix and chapter 3 table names now

i think table 9.1 also could have a label like this too
It is labelled in 
http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#discrete-sampling-geometries
but at the bottom of the table

Do you think we should maintain internal consistency (add a change) or consistency with the 1.6 layout (no change to this PR)?

rhattersley:
> Do you think we should maintain internal consistency (add a change) or consistency with the 1.6 layout (no change to this PR)?

The approach that @davidhassell & I have settled on so far has been to mimic 1.6 as far as possible. So although it's weird & inconsistent that the tables in chapter 9 don't have "proper" labels I've left fixing that for the next version.

marqh:
that's fine, i think there are multiple instances in any case, not just 9.1

let's agree on consistency for now

PR #42, opened 2015-11-22
Consistent column widths for chapter 9.

rhattersley:


marqh:
i can't see any difference in the output from this PR

am i missing something?
what should be the result of this change please?

rhattersley:
Currently [table 9.4](http://cf-metadata.github.io/#_contiguous_ragged_array_representation) is full width. With the change it should only occupy 25% of the width, thus the cells end up the same width as in earlier 4-column tables.

This gives consistency across the various tables, and also makes it look much more like the original DocBook version.

marqh:
that makes sense, thank you

PR #45, opened 2015-11-23
Preface fixes

### LINKED ISSUE #14: Preface revision history
In the preface, the existing description of the revision history is no longer accurate.

In particular, if we are no longer using the pink, yellow, green highlighting scheme the following phrase no longer applies:

> Changes with provisional status use the following mark-up style: ...

Also, we need to consider if the following phrase is still accurate:

> See Appendix G, Revision History for the full revision history. 

Perhaps the following wording would be more appropriate?

> This document will be updated to reflect agreed changes to the standard and to correct mistakes according to the rules of [CF governance](http://cfconventions.org/governance.html). See [Revision History](http://cf-metadata.github.io/#revhistory) for the revision history prior to the move to GitHub. See [GitHub](https://github.com/cf-convention) for the subsequent revision history.

rhattersley:
Tidy-up syntax and clarify revision history status.

Closes #14.

### DIFF
--- a/pr01.adoc
+++ b/pr01.adoc
@@ -1,17 +1,16 @@
-
 [[preface]]
 [preface]
-[]
 == Preface
 
-Home page::: Contains links to: previous draft and current working draft documents; applications for processing CF conforming files; email list for discussion about interpretation, clarification, and proposals for changes or extensions to the current conventions. link:$$http://www-pcmdi.llnl.gov/cf/$$[http://www-pcmdi.llnl.gov/cf/] 
+Home page: ::
+Contains links to: previous draft and current working draft documents;
+applications for processing CF conforming files; email list for
+discussion about interpretation, clarification, and proposals for
+changes or extensions to the current conventions.
+link:$$http://www-pcmdi.llnl.gov/cf/$$[http://www-pcmdi.llnl.gov/cf/]
 
-Revision history::: This document will be updated to reflect agreed changes to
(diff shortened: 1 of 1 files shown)


rhattersley:
@JonathanGregory, @davidhassell, @rsignell-usgs, @marqh - if we can merge this PR and the other recent PRs I've raised we will have addressed all known issues ... exciting times!

marqh:
> exciting times!

exciting indeed

this one is fine, it now matches the current 1.6 doc

rhattersley:
Thanks @marqh

JonathanGregory:
It looks lovely - thanks very much and well done. Is there a way to make a PDF of it, so I can print it out nicely and sit down with it and the existing PDF to compare them side-by-side?

I noticed one thing in the preface: the link is out of date. http://www-pcmdi.llnl.gov/cf/ should be replaced with http://cfconventions.org/.

Best wishes

Jonathan

rhattersley:
> I noticed one thing in the preface: the link is out of date

Yes - I spotted that but left it alone to respect the existing DocBook version. If you're happy for it to be updated then I can apply the quick fix. Or you might like to click on the pencil icon for [pr01.adoc](https://github.com/cf-metadata/cf-conventions/blob/master/pr01.adoc) (with the tool-tip text "Fork this project and edit the file") which will let you modify the text and create a new pull request [all in the browser](https://help.github.com/articles/editing-files-in-another-user-s-repository/).

PR #46, opened 2015-11-25
Fix reference to COARDS.

rhattersley:
As can be seen with all the other bibliography references, the square brackets come for free now.

### DIFF
--- a/appa.adoc
+++ b/appa.adoc
@@ -191,7 +191,7 @@ Attribute
 | **`positive`**
 | S
 | C
-| [<<COARDS>>]
+| <<COARDS>>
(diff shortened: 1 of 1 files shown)


rhattersley:
Rebased.

marqh:
this one empties the 1.6 milestone

I suggest we consider a 1.6 tag

rhattersley:
> I suggest we consider a 1.6 tag

Nearly ... #48 is still open. #49 is a big step to addressing that but we need to give people (e.g. @JonathanGregory) time to look over the PDF output too.

The tag shouldn't be created until the HTML and PDF outputs are agreed to be ready for publishing via the main website.

PR #47, opened 2015-11-25
correct the CF link

JonathanGregory:
Dear Richard

I respect your respecting the existing text! I think that's the right thing to do regarding the conventions, but since this new version may well supersede the existing 1.6 document, or be maintained together with it, I think it makes sense to correct the link.

Best wishes

Jonathan


rhattersley:
Perfect :+1: 

Thanks for the fix. :smile:

Issue #48, opened 2015-11-25
PDF output


rhattersley:
See https://github.com/cf-metadata/cf-conventions/pull/45#issuecomment-159609859

marqh:
I think this is done now
http://cf-metadata.github.io/cf-conventions.pdf

marqh:
this is the last open issue in 
https://github.com/cf-metadata/cf-conventions/milestones/1.6
i wonder if our goalposts are still minded to grow legs
:smirk:

rhattersley:
I believe @JonathanGregory is intending to review the PDF output as well, so I'd like to hold fire for now.

JonathanGregory:
Yes, I am going to review it. Thanks. I can't justify that much paper so I will look at the two PDFs adjacently online. Jonathan

marqh:
I believe that all active pull requests in the 1.6 milestone are now closed.

It is my view that this issue is now resolved.

@rhattersley @JonathanGregory 
do you have any reasons to keep this ticket open, or may I close it?

Would you like to include any deployment tickets in the 1.6 milestone or may we also conclude that as achieved?

thank you
mark

JonathanGregory:
How marvellous. Well done on this remarkable achievement.

If it's now ready to go live, as HTML and PDF, I think we should just check with someone first. Shall I ask Karl if that's OK?

Jonathan

rhattersley:
> Shall I ask Karl if that's OK?

That sounds like a super plan. :grinning:

rhattersley:
For the record, Karl is on-board, but has also requested Jeff's opinion.

rhattersley:
Since the remaining room for debate seems to be more about "how do we make v1.7 from here?", rather than any issues with the current version, I've tagged [v1.6](https://github.com/cf-metadata/cf-conventions/releases/tag/v1.6.0). :grinning:

It seems fair to close this issue now. It's easy to open new ones!

PR #49, opened 2015-11-25
Use built-in entity attributes for curly quotes.

rhattersley:
This fixes the parsing failures when producing a PDF with asciidoctor-pdf.

It also turns on automatic building of the PDF by Travis-CI... although that won't be apparent until this PR is merged.

Relates to #48.


rhattersley:
NB. I've manually put up a preview of the results of merging this PR: http://cf-metadata.github.io/cf-conventions.pdf

rhattersley:
I was concerned that the [new PDF version](http://cf-metadata.github.io/cf-conventions.pdf) had some page breaks in undesirable places, but I've just taken another look at the [old PDF version](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.pdf) and it suffers from the same sorts of problems. For example, table 3.1 is split in the old version, but table 3.3. is split in the new version. So now I'm more inclined to think the quality is comparable.

marqh:
looks good to me.  I think the markup changes make the source easier to read, which is grand.

the build works for me locally for pdf and html

i haven't checked the travis build, that can run once the PR is up and we can catch any issues if they arise

PR #52, opened 2015-11-29
Turn on numbering for Appendix H.

rhattersley:
Unlike the other appendices, appendix H does use numbers for the section & sub-sections.


marqh:
agreed, inconsistent, but absolutely reproducing the content

PR #53, opened 2015-11-29
Fix text for NUG links.

rhattersley:
Slightly more controversial ... this actually updates the text of several links to reflect reality (as it currently stands).

### DIFF
--- a/appa.adoc
+++ b/appa.adoc
@@ -22,7 +22,7 @@ Attribute
 | **`add_offset`**
 | N
 | D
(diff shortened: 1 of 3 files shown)


marqh:
@JonathanGregory do you agree that this update should be put into the 1.6 document managed here, despite differences with 
http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#attribute-appendix
?

e.g. the URLs in the table links are the same, but the label is different

JonathanGregory:
Yes - I suggested it. :-) The document we link appears to have changed its arrangement, but it looks like the right content to me i.e. the attribute list is not NUG 8.1 any more, but Appendix B. If you agree, I think this change is fine to make.

marqh:
that's fine with me, thank you

PR #54, opened 2015-11-29
Update URLs for standard name table.

rhattersley:


rhattersley:
This could do with an update to the URL for standard region names.

marqh:
> This could do with an update to the URL for standard region names.

this one
http://cfconventions.org/Data/cf-standard-names/docs/standardized-region-names.html
?

should this include a link to area_types as well?
http://cfconventions.org/Data/area-type-table/3/src/area-type-table.xml

I don't think these links are in the current conventions doc, but I support having them as resolvable links in this version

is this something I should be wary of agreeing to change?

JonathanGregory:
I agree that would be a sensible thing to do, but adding a new link would be a change to content (although a minor and one) rather than fixing a defect (e.g. in formatting, out-of-date text or URLs), I would say. If you open a ticket to do this, I'll support it. Jonathan

rhattersley:
I'm only proposing to update the link in [6.1.1 Geographic Regions](http://cf-metadata.github.io/#geographic-regions) to the "standardized region names". I'm not proposing any change of content.

I'll post the actual change soon to make it clear.

JonathanGregory:
OK, updating the link would be fine. Jonathan

rhattersley:
Thanks for being patient. I've updated the region URL now, so it's ready for feedback/merge as appropriate.

marqh:
this looks grand: all the URLs resolve as expected to the reqiured content

cheers

PR #56, opened 2015-11-29
Move table of contents to start of document.

rhattersley:
It's not yet possible to specify the position of the table of contents when producing the PDF document. So instead, this PR modifies the structure of the document to fit the fixed location of the table of contents.
1. Allows the HTML & PDF versions to have the same structure.
2. Allows the lists of tables & examples to follow the table of contents in the PDF document.

Previews of the new structure are available at:
- http://cf-metadata.github.io/moved-toc.html
- http://cf-metadata.github.io/moved-toc.pdf

It's not clear whether this new structure is an improvement. Feedback welcome!

### DIFF
--- a/cf-conventions.adoc
+++ b/cf-conventions.adoc
@@ -5,6 +5,11 @@ Version 1.6, 5 December, 2011
 :toc: macro
 :toclevels: 3
 
+toc::[]
+
+:numbered!:
+include::toc-extra.adoc[]
+
 == About the authors
 
 .Original Authors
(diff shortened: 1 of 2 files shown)


marqh:
this is better from my point of view, as well as being consistent in html and pdf

I'll leave a little time for other comments before adoption

JonathanGregory:
Thanks. I think your contents workround is sensible and looks satisfactory.

marqh:
grand, let's go with this then

PR #64, opened 2016-01-13
requirements recommendations

marqh:
I have added the conformance page (as html) to this repository

I think the conformance document and the conventions document are aimed to be managed in step, so this makes sense to me.

I have added the 1.6 snapshot to be consistent with the conventions content version.  
If this is adopted, we can bring in the 1.7 draft, and thus track changes

If a different format is preferred, please may content in this form be provided? The html form
was all I could find.

thank you
mark

### DIFF
--- a/requirements-recommendations.html
+++ b/requirements-recommendations.html
@@ -1,0 +1,287 @@
+<h1 class="documentFirstHeading">CF Conformance Requirements and Recommendations 1.6</h1>
+<div class="plain">
+  <p>&nbsp;</p>
+  <ul><li>The following is a list of requirements and recommendations for a CF
+    conforming netCDF file.  They are organized by the section of the CF
+    document that they pertain to.</li></ul>
+  <ul><li>This document is intended to be a concise summary of the <a href="/cf-conventions/v1.6.0/cf-conventions.html">CF conventions 1.6</a> document.  If there are any
+    discrepencies between the two, the conventions document is the ultimate
+    authority.
+    </li></ul>
+    <ul><li>This document will updated as required to correct mistakes or add new
+      material required for completeness or clarity.</li></ul>
+    <p>&nbsp;</p>
+    <h4>2.1 Filename</h4>
+    <strong>Requirements:</strong>
+    <ul><li>Filename must have ".nc" suffix.</li></ul>
+    <h4><br /></h4>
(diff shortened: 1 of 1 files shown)


painter1:
I have to apologize for not noticing this pull request.  Last week I took the html version of the draft 1.7 conformance document and converted it to AsciiDoc (mostly; lots of variable names still have to be highlighted).  I put it in a separate repository, named cf-convention/Conformance.

I don't have any strong opinions about whether the conformance document should be in the same repository as the main CF Conventions document, or its own repository.

One thing which I am concerned about relates to a feature we have for the main document.  We have a nice system for automatically converting the master branch's AsciiDoc files into an html file, which is immediately visible to the community.  Recently this worked as it should for Ticket 145.  I would like to see the same sort of system for the conformance document, but I don't know how it works or who originally set it up.

rsignell-usgs:
I know that everything that happens automatically upon PR or merge is controlled by 
https://github.com/cf-convention/cf-conventions/blob/master/.travis.yml
but I don't know much more than that.   

I think unfortunately @rhattersley is no longer involved with CF, but perhaps someone else from the @scitools group or @ocefpaf would be willing to help us out?

rsignell-usgs:
Maybe it's as simple as adding a few lines here?
https://github.com/cf-convention/cf-conventions/blob/master/.travis.yml
@painter1 , what would those lines be?

ocefpaf:
> https://github.com/cf-convention/cf-conventions/blob/master/.travis.yml#L21

I did not check the details but I do see all the parts necessary for an automatic update of the page every time a PR is merged in https://github.com/cf-convention/cf-conventions/blob/master/.travis.yml#L21

(Although nowadays I would re-write that to use `doctr` and simplify it a lot.)

rsignell-usgs:
@painter1, ah, okay, I see you put the conformance doc in it's own repo.   So you would need to add a .travis file similar to https://github.com/cf-convention/cf-conventions/blob/master/.travis.yml and get the tokens set up there.

dblodgett-usgs:
This PR is getting very long in the tooth. Can we agree to merge or close it for later?

dblodgett-usgs:
Going to close and not merge. We need to figure out how to get these builds working properly.

PR #66, opened 2016-01-20
Apply changes from trac ticket 67.

rhattersley:
See http://cf-trac.llnl.gov/trac/ticket/67

### DIFF
--- a/appa.adoc
+++ b/appa.adoc
@@ -180,7 +180,7 @@ Attribute
 | D
 | C, D
 | <<missing-data>>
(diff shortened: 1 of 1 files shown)


marqh:
consistent with hash67

rhattersley:
@marqh - you might want to hold fire for a bit. I think my PRs need to include updates to the "Revision history".

marqh:
apologies: would you like me to revert, or is a follow up PR useful.

I confess I think that this a small editorial change, not requiring a 'Revision History' entry, but i am very easily swayed from that point of view

rhattersley:
Other similarly small defects have their own entry in the revision history, so since we're sticking with that way of recording changes I think it's needed here too.

I don't a revert is needed - I'll handle it in a follow-up.

PR #69, opened 2016-01-20
Apply changes from trac ticket 143.

rhattersley:
See http://cf-trac.llnl.gov/trac/ticket/143

NB. This does not include the corresponding changes to the conformance document. See #64 for a proposal to include the conformance document in this repo.
- [x] Added link from trac ticket to this PR?
- [ ] Updated "Revision History"? (Use the date you applied the changes.)


painter1:
The new Appendix A contains the text "FIXME: "parameterized dimensional vertical coordinate" …​ obsolete terminology?".    In the context of the ticket, it doesn't look obsolete to me.

painter1:
I'm sorry, I was looking at the wrong version of section 4.3.3 when I wrote the previous comment.  I agree that "parameterized dimensional coordinate values" should be reworded to agree with section 4.3.3, perhaps to "computed vertical coordinate values" as in the final version of section 4.3.3.

painter1:
None of the standard names in the proposal have been approved yet, according to http://cfconventions.org/Data/cf-standard-names/40/build/cf-standard-name-table.html

JonathanGregory:
I have asked Alison re stdnames cc Jeff.

painter1:
In a moment I will merge this with git command line tools..

I have resolved the "FIXME" issue in Appendix A (the row for computed_standard_name) by using "computed vertical coordinate values" rather than "parameterized dimensional vertical coordinate values" as in the original proposal of ticket 143.  As @rhattersley pointed out, the proposal's original terminology for this row is not consistent with the final terminology of comment 9.   Also I have changed the title of Appendix D to "Computed Vertical Coordinates" as implied by the body of comment 9.

I have not changed the title of the new section 4.3.3.

JonathanGregory:
Dear Jeff
In comment 8 of the ticket I suggested "Parametric vertical coordinates" and that seemed to be acceptable. That should therefore be the title of the new section 4.3.3 and of Appendix D, and "parameterized" should be replaced with "parametric" wherever it occurs. I agree with your  used of "computed" in the text of the new row in Appendix A.
Thanks
Jonathan

painter1:
The term "parameterized" only appears in the titles of section 4.3.3 and Appendix D.   I changed both titles to use "parametric", and the cross-references to them.

I did this directly in the master branch because it is a minor change and already vetted.

davidhassell:
Looks good - thanks.
David

PR #74, opened 2016-03-17
Revision history entry for trac ticket 67.

rhattersley:
A follow-up to #66.


rhattersley:
Ping @marqh

marqh:
that's great, thank you

PR #79, opened 2017-01-25
123

reshel3:
Amended sentence in first para of Sect 3.3


mattben:
👍

painter1:
This is exactly the change needed.  But we also need a brief entry into the revision  history.

painter1:
I've figured out how to update the revision history myself, using shell commands.

PR #80, opened 2017-01-25
103 Appex a

reshel3:
Removed "C" from "cf_role"

### DIFF
--- a/appa.adoc
+++ b/appa.adoc
@@ -69,6 +69,7 @@ Attribute
 
 | **`cf_role`**
 | S
+| C
 | <<coordinates-metadata>>
 | Identifies the roles of variables that identify features in discrete sampling geometries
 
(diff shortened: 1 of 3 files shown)


painter1:
This pull request addresses ticket #123 exactly as it should.  It also renames Appendix G to Appendix Z.  I think that this change to ticket #123 should be approved in the other place where it appears, pull request #79, titled "123", once it comes with an entry in the revision history. 

At that point we can decide whether it is convenient to rename the appendix for the moment.  Eventually it will have to be changed to an unnumbered chapter like the bibliography.

painter1:
I can't find the changes to Appendix H here.  It looks as though the Appendix A changes may already have gotten into the master branch somehow, though I haven't checked them all.  Finally, the master branch has been changed since this branch was produced.   Overall, I think the easiest solution may be to abandon this branch and do the Appendix H changes in a new branch, branched off of the current master branch.

PR #73, opened 2016-03-04
Create PULL_REQUEST_TEMPLATE.md

rhattersley:
Adds a [pull request template](https://github.com/blog/2111-issue-and-pull-request-templates) to help ensure traceability and appropriate revision history updates.

PR #81, opened 2017-01-26
149

reshel3:
Changed current standard_name from "area" to "cell_area" in Section 7.3

### DIFF
--- a/appa.adoc
+++ b/appa.adoc
@@ -69,6 +69,7 @@ Attribute
 
 | **`cf_role`**
 | S
+| C
 | <<coordinates-metadata>>
 | Identifies the roles of variables that identify features in discrete sampling geometries
 
--- a/appa.adoc
+++ b/appa.adoc
@@ -69,7 +69,6 @@ Attribute
 
 | **`cf_role`**
 | S
-| C
 | <<coordinates-metadata>>
 | Identifies the roles of variables that identify features in discrete sampling geometries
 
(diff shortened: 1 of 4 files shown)


painter1:
This branch has the defect corrections requested in tickets 123 (chapter 3) and 149 (chapter 7).  These changes are exactly correct.   But also these revisions need a one-line description in the revision history, appendix G.  And the tag for appendix G, at the top of the file, should be changed from Z back to G as it was before.

painter1:
This pull request should not be merged until pull request #79 titled "123" has been merged.  At that point this pull request will be only about issue #149, and appendix renaming.  Then this pull request will need an entry into the Revision History (Appendix G or Appendix Z), briefly stating that it resolves ticket #149.

There will be no need to mention the appendix renaming in the Revision History, because we know that it doesn't fully resolve ticket #73.  When ticket #73 is fully addressed, then we can say so in the Revision History.

painter1:
I wouldn't link to cf-trac because we may change the url.  The ticket number is necessary and sufficient.  If you think that a brief description of the change would be a helpful addition, feel free!

I think that it doesn't matter much whether changes are listed together or separately.  The only problem with listing them together would arise if not all the described changes actually made it into the document, i.e. if some problem postponed a pull request merger.   Then the revision history would have to be retroactively changed - not a big deal, but some extra trouble.

painter1:
merged, resolving conflicts and updating the Revision History, using command-line tools

PR #82, opened 2017-01-26
100

reshel3:
Replaced text in Para 1, 2, 3, 5 and appended text to para 5.


JonathanGregory:
Dear Tanya

It looks to me that ticket 100 hasn't been completely implemented here. It's not exactly as it was stated at the start of the ticket, because of subsequent discussion. See comments 3 and 8 in the ticket.

Best wishes

Jonathan

painter1:
Tanya's changes were complete as of today.  I merged them with command-line tools, after adding or cleaning up a few cross-references.

PR #83, opened 2017-01-30
71

reshel3:
71 - Added text to Vetical Perspective

### DIFF
--- a/appf.adoc
+++ b/appf.adoc
@@ -285,8 +285,7 @@ __Notes:__:: A general description of vertical perspective projection is
 given in <<Snyder>>, pages 169-181.
 
 +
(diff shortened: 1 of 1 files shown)


painter1:
This ticket has already been done by Richard Hattersley, c.f. comment 4.  The change here, to appf.doc, is to replace "PROJ.4" with "proj.4" (also two lines are merged into one, but that has no effect after processing by ASCIIDOC).

The lower-case name is indeed used in the ticket.  But the official web site, proj4.org, uses it both ways - e.g. lower-case in the home page, upper-case in the FAQ page.   So either upper- or lower- case is ok.   Because upper-case is used in the rest of Appendix F, I'd be inclined to leave it as upper-case here too.

PR #84, opened 2017-01-30
87

reshel3:
87 - Changed in Appendix A,  "Comment" attribute from G, D to G, C, D

### DIFF
--- a/appa.adoc
+++ b/appa.adoc
@@ -81,7 +81,7 @@ Attribute
 
 | **`comment`**
 | S
-| G, D
+| G, C, D
 | <<description-of-file-contents>>
 | Miscellaneous information about the data or methods used to produce it.
 


painter1:
What's the ticket number for this change?

painter1:
I'm sorry, I see that the ticket number is the title of this pull request.

Anyway, the change is exactly correct.  All we need is an entry in the Revision History.

davidhassell:
Who does have write access? I don't think I do! I presume that Richard Hattersley does...anyone else?

painter1:
Write access to what?  Probably you want to be on the "Devel Team".  It has 21 members.  I'll send you an invitation.  Probably that will do what you need.

painter1:
David, you have already been invited to the devel team.  I don't see a way to re-send the invitation, so if you can't find it, please let me know.  Then I'll cancel the invitation and create a new one.

painter1:
updated revision history and merged from the command line

PR #85, opened 2017-02-02
#77

reshel3:
#77 Added Sinusoidal as a grid_mapping_name

### DIFF
--- a/appf.adoc
+++ b/appf.adoc
@@ -220,6 +220,24 @@ __Notes:__::
 {nbsp}
 
 
+=== Sinusoidal
+
+----
+
+grid_mapping_name = sinusoidal
+        
+----
+
+__Map parameters:__::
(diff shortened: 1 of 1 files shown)


painter1:
This change is correct and only needs an entry in the Revision History.

It's worth nothing that the link to Snyder doesn't jump to the bibliography.  None of the other ones in this appendix, Appendix F, work any better, so I don't think that this is an essential change.  I think that something like <<bibliography.adoc#Snyder>> might work better than <<Snyder>> but I haven't tested that ides.

davidhassell:
+1 for #77 changes. Do the citation links work from non-appendices?

painter1:
I fixed the references to Snyder.  Only one had to be changed to make the other two work.
I also updated the Revision History, then merged from the command line.

PR #86, opened 2017-02-02
#92

reshel3:
#92 Added Oblique Mercator as a grid_mapping_name and added
azimuth_of_central_line as an attribute


painter1:
Everything is done right.  In particular, the revisions from the discussion are all correctly here.  This only needs an entry in the Revision History.

davidhassell:
#92 looks good to me, too.

painter1:
I merged with command line tools, after updating the Revision History and making two minor changes:  put grid_mapping_name in lower case like the other projections; and replaced the title's underscore with a space, also like the other projections.

PR #72, opened 2016-02-29
Apply changes from trac ticket 103.

mo-g:
See http://cf-trac.llnl.gov/trac/ticket/103

@rhattersley Opinion of my interpretation?

PR #87, opened 2017-02-06
#72

reshel3:
#72 Added Geostationary projection to grid-mapping in Appendix F

### DIFF
--- a/appf.adoc
+++ b/appf.adoc
@@ -59,6 +59,32 @@ __Map coordinates:__:: The x (abscissa) and y (ordinate) rectangular coordinates
 __Notes:__:: Notes on using the **`PROJ.4`** software package for computing the mapping may be found at link:$$http://www.remotesensing.org/geotiff/proj_list/azimuthal_equidistant.html$$[http://www.remotesensing.org/geotiff/proj_list/azimuthal_equidistant.html].
 
 
+=== Geostationary projection
+
+----
+
+grid_mapping_name = geostationary
+        
+----
+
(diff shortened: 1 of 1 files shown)


painter1:
The Notes section has some formatting problems.
None of the urls work any more, although they must have at the time that ticket #72 was discussed and this revision defined.  I'll try to find the currently correct ones.
Of course, the Revision History will need an entry for this.

JonathanGregory:
Martin Raspaud should be named in the list of additional authors of the CF convention, as mentioned in comment 19 on this ticket.

davidhassell:
I've had a little dig around on the internet, and a had a reread of http://cf-trac.llnl.gov/trac/ticket/72, and I'm fairly happy that the three broken urls:

http://www.eumetsat.int/idcplg?IdcService=GET_FILE&dDocName=PDF_CGMS_03&RevisionSelectionMethod=LatestReleased
​http://trac.osgeo.org/proj/wiki/proj%3Dgeo
​http://remotesensing.org/geotiff/proj_list/geos.html

Could now be:

www.cgms-info.org/documents/pdf_cgms_03.pdf
http://proj4.org/
https://trac.osgeo.org/geotiff/

What do you think?

painter1:
Martin Raspaud got into the list of additional authors in another branch was was merged in earlier; I'm not sure which one.  @davidhassell's new urls are in, thanks to Tanya.  I fixed the formatting errors in the Notes section and updated the revision history, then merged with command-line tools.

PR #88, opened 2017-02-13
#118

reshel3:
#118, Add geoid_name and geopotential_datum_name to the list of
Attributes.


JonathanGregory:
That's fine, thanks. As with all tickets, it needs a comment in the revision history too. I suppose it would make sense to put those updates in the same pull request, wouldn't it? Jonathan

PR #89, opened 2017-02-13
#148

reshel3:
#148, Added maximum_absolute_value, minimum_absolute_value and
mean_absolute_value to cell methods in Appendix E


JonathanGregory:
Thanks. Looking at the diff, I notice something I hadn't before: I think that in alphabetical order maximum_absolute_value should come after plain maximum, and similarly for the others. It's a small point, but would be an improvement I think. Jonathan

painter1:
merged with command-line tools

PR #75, opened 2016-03-30
Changed affiliations of Jonathan Gregory and Phil Bentley.

mo-g:
See http://cf-trac.llnl.gov/trac/ticket/141

 - [ ] Added link from trac ticket to this PR?

 > Applied via ​https://github.com/cf-convention/cf-conventions/pull/XXX

 - [X] Updated "Revision History"? (Use the date you applied the changes.)

PR #90, opened 2017-02-16
Changes required by ticket #75

painter1:
This has the changes specified in ticket 75, except for the urls and related text.
Please do not merge this pull request until we have correct urls, and surrounding text.

The ticket's urls are now invalid.  For lambert_conformal_conic, I replaced the two urls with a more generic reference to http://proj4.org/projections.  For polar_stereographic, I have made no replacement.  Proj4.org lists "universal polar stereographic" which I think might be slightly different from "polar stereographic"; I need more advice here.

For all grid mappings, the grid-mapping-specific text at proj4.org is extremely terse, briefer than the references to it in Appendix F!

### DIFF
--- a/appf.adoc
+++ b/appf.adoc
@@ -135,6 +135,9 @@ __Map coordinates:__:: The x (abscissa) and y (ordinate) rectangular coordinates
 __Notes:__:: Notes on using the **`PROJ.4`** software package for computing the mapping may be found at link:$$http://proj4.org/projections/lcc.html$$[http://proj4.org/projections/lcc.html].
 and
 link:$$http://geotiff.maptools.org/proj_list/lambert_conic_conformal_1sp.html$$[http://geotiff.maptools.org/proj_list/lambert_conic_conformal_1sp.html]
+("Lambert Conic Conformal (1SP)" or EPSG 9801) or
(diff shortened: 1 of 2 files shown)


davidhassell:
[4c5ba55](https://github.com/cf-convention/cf-conventions/pull/90/commits/4c5ba5588e8e149371584bdbfb4eb3c230c1dfa9) Looks good. Thanks.

painter1:
@davidhassell , what url would you use for how to use PROJ.4 for computing a polar stereographic projection?

davidhassell:
For polar steographic http://geotiff.maptools.org/proj_list/polar_stereographic.html looks good. It also include the transformation formulae, too.

davidhassell:
Perhaps we should use http://geotiff.maptools.org/proj_list/PROJECTION_NAME for all of the grid mappings, in addition to  http://proj4.org/projections ?

painter1:
OK, I'll update all the dead urls in Appendix F.

painter1:
I believe that all changes for issue 75 are implemented correctly now.

Also all links in Appendix F are live.  The ones for "geostationary projection" may not be relevant, but I have no better ones ("geostationary satellite view" looks different).  In a couple places there is wording involving "EPSG" which may or may not refer to contents of now-dead web pages.  I left them alone.

PR #91, opened 2017-02-16
#103

reshel3:
#103 FillValue for the variables humidity and temp should be -999.9f
added "f" as requested


painter1:
Merged with command-line tools.

PR #97, opened 2017-03-21
Reshel 80

painter1:
See https://cf-trac.llnl.gov/trac/ticket/80.

There are still some obsolete urls; do not merge this pull request until they are corrected.

But for now, I would like some others to look at these changes seriously.  Tickets 69 and 80 overlap in section 5.6 and the new 5.6.1.  In consequence, implementing ticket 80 (after 69 was already in there) required some substantial rearrangements, and resolution of a conflict or two.  This almost entirely pertains to the examples now numbered 5.9, 5.10, 5.11, and 5.12.  I tried to touch things as little as possible, but I would not be surprised if someone asks for a correction or further work.


painter1:
I have finished fixing the urls.  That includes moving the documents they reference to the github wiki; one of them no longer existed on the web other than at archive.org.  The links within the documents have also been fixed.

However, we should not merge this branch until we have given other interested parties enough time to review it and comment on it.

davidhassell:
I've read chapter 5 and have a couple of points:

* The line _"use of the TOWGS84 element to specify horizontal datum transformation parameters (to WGS 1984 datum)"_ is not needed, because there is now a CF grid mapping attribute (`towgs84`)

* I'm not sure where the line _"Also for clarity, we have dropped the quotation marks which would delimit the entire crs_wkt string"_ comes from (it's not in tickets 69 nor 80 as far as I can tell). Nonetheless, it's a good point, but should, perhaps be rolled into the this paragraph: _"Note: To enhance readability the WKT value has been <snip> mandated by the CF convention"_, which comes from ticket 69.

Thanks, David

painter1:
@davidhassell , regarding your TOWGS84 point:  To clarify, do you mean to drop that line of text without changing the examples?  That would make sense to me because crs_wkt is only a supplement to other attributes and redundancy is explicitly permitted.

The line about quotation marks was mine - I felt I would be changing things less if I explained why the examples were incorrect than if I corrected them myself.  You are right that the note on the bottom is a much better place for it.

I have made both changes now.

davidhassell:
@painter1 , These changes look fine to me - thanks. Re. TOWGS84, yes, I agree - drop the line in the list but leave the examples unchanged.

PR #108, opened 2017-04-06
Ticket 70, Connecting coordinates to Grid Mapping variables

painter1:
Chapter 5 is revised as specified in Ticket 70.

However, the consequent new Example 5.10 is no longer consistent with Example 5.12.  So I also merged the old Example 5.10 into Example 5.12, and made the necessary small changes in the associated text.


davidhassell:
Hi,

Minor point: Would it better to italicise the first (rather than 2nd) occurrence of "grid mapping variable"?

I like your rewordings from the ticket - much clearer.

David

painter1:
@davidhassell, thanks for catching the italics placement issue.  I've fixed it.

PR #116, opened 2017-06-29
Ticket #140

reshel3:
Ticket #140, Added 3 paragraphs and an example to Chapter 7, Section
7.1.


davidhassell:
Looks good - thanks, Tanya.

painter1:
It looks good to me too, except that it needs two minor changes:  (1) "====" is missing at the end of the new example, and this messes up the following formatting and section numbering; and (2) the examples should be renumbered because there is a new one.  In the next few minutes I will make those corrections and merge them in.

PR #118, opened 2017-07-10
New Table D.1 format

painter1:
I replaced Table D.1 with one formatted as requested by Karl Taylor.

### DIFF
--- a/appd.adoc
+++ b/appd.adoc
@@ -462,28 +462,36 @@ The `standard_name` for `depth` and the `computed_standard_name` must be one of
 the consistent sets shown in <<table-computed-standard-names, Table D.1>>. No
 `standard_name` has been defined for `z1`, `z2`, `a`, `href` or `k_c`.
 
-
+// This table has an unusually long title, and AsciiDoc is unable to wrap it.
+// The best I can do is to use two titles, one for an empty table...
 [[table-computed-standard-names]]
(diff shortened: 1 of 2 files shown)


painter1:
The main problem I had was that the title is long, and AsciiDoc cannot wrap a block title.  My solution was to use half the title for one, blank, table and the other half of the title for the real table - rather kludgy.

painter1:
Actually GitHub's AsciiDoc engine seems to be able to wrap a block title.  It's asciidoctor that can't.

taylor13:
Hi Jeff,

This looks good to me.  I think the title would be clearer if it began:  "Consistent sets of values for the standard_names of formula terms and the computed_standard_name ... "  If you agree, why don't you go ahead and change it and call it done.
thanks,
Karl

painter1:
I've changed the title.  As I mentioned elsewhere, my workaround to Asciidoctor limitations leads to a harmless but ugly horizontal line in the title.  But there's no better alternative which I can find.  So I also think we're done now.

painter1:
I solved that title problem.  I did it working directly in the master branch because I had to test it with asciidoctor-pdf as well as asciidoctor, and a firewall problem has kept me from installing asciidoctor-pdf.

It turns out that asciidoctor and asciidoctor-pdf _will_ wrap the title of a figure.  So I put the entire title on a figure, right before the table.  The image displayed is 1 pixel and transparent, hence invisible.  So it looks like the table's title.

PR #120, opened 2017-07-19
use up to date OGC reference to WKT-CRS

marqh:
See http://cf-trac.llnl.gov/trac/ticket/162

 - [x] Added link from trac ticket to this PR?

 > Applied via ​https://github.com/cf-convention/cf-conventions/pull/XXX

 - [ ] Updated "Revision History"? (Use the date you applied the changes.)

I'm aware this is late to the party.  I have raised a Trac ticket regarding this, I wanted to capture the details inside github as well

@painter1 where are we in the time line for adopting CF1.7 please?
I can see from
https://github.com/cf-convention/cf-conventions/pull/119
that 'master' is the 'Final draft'

is master still accepting editorial/defect changes?
is master accepting material changes?

clearly we'll want a view from the community on whether this is material or editorial

many thanks
mark

### DIFF
--- a/appf.adoc
+++ b/appf.adoc
@@ -417,7 +417,7 @@ to a reference direction.
 | **`crs_wkt`** | S
 | This optional attribute may be used to specify multiple coordinate system properties
   in well-known text (WKT) format. The syntax must conform to the WKT format as specified
-  in reference <<OGC_CTS>>. Use of the **`crs_wkt`** attribute is described in section 5.6.1. 
+  in reference <<OGC_WKT-CRS>>. Use of the **`crs_wkt`** attribute is described in section 5.6.1. 
 
 | **`earth_radius`** | N
 | Used to specify the radius, in metres, of the spherical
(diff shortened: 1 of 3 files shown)


davidhassell:
I think that changing the URL to something that someone who's in the know thinks is better is editorial at this stage

davidhassell:
Can we change ​http://www.opengeospatial.org/standards/ct to ​http://www.opengeospatial.org/standards/wkt-crs before release? I don't think that it needs further discussion as it has had the same rigour applied to it as other URLs. 

I don't think that we need to wait the customary 3 weeks for ticket #162 on trac, either, in this case.

Thanks and all the best,

David

painter1:
Mark, the changes are in your branch - if you agree with David's change, could you put it in?  I'll be ready to merge the pull request once this is all settled.

This is just under the wire!  Last week I was about to change the web site to point to 1.7, and announce it on cf-metadata.

painter1:
Now that I have actually looked at the document, I see that I had misinterpreted what David said in his note.  There are no more url changes.  However, in the Trac system for issue 162, I have a question regarding a statement about Example 5.12.

painter1:
I have merged the pull request.  I haven't closed ticket 162, however, because we may still need a small further change.

PR #149, opened 2018-11-05
move CONTRIBUTING out of .github dir

ChrisBarker-NOAA:
CONTIRBUTING.md needs to be easy to find -- I knew it was there somewhere, and I couldn't find it. I only found it by finding the issue where is was discussed, then the link to the PR, then looking at the PR.

.github is a hidden dir on many systems, not a good place to put a file yu want folks to find up front!

Note that gitHub suggests:

  To help your project contributors do good work, you can add a file with contribution guidelines to your project repository's root, docs, or .github folder. When someone opens a pull request or creates an issue, they will see a link to that file

So I can see how it got there, but it can be in root too, where it is far more likely to be found.


dblodgett-usgs:
I think I was following old guidelines by putting it in .github. I'm glad you made this suggestion. I support this PR.

davidhassell:
Good idea, Chris. I also support this.

Thanks, David

dblodgett-usgs:
Given that this is a non-controversial change and does not apply to the specification, I'd say we should merge given we have two people in support. Any dissent?

ChrisBarker-NOAA:
Not me, obviously:-)

davidhassell:
That's good for me, too.

Issue #156, opened 2019-03-01
Alternate grid mappings for geometry containers


twhiteaker:
**Title:** Alternate grid mappings for geometry containers
**Moderator:** @davidhassell 
**Requirement Summary:** 
Allow geometries to have a grid_mapping different than the data variable. For example, data variable coordinates could be in geographic coordinates, while coordinates for watershed polygons could be in a projected coordinate system.

**Technical Proposal Summary:** 
Allow the `grid_mapping` attribute value on the geometry container to differ from the `grid_mapping` attribute on the data variable, provided the auxiliary coordinates do not have the nodes as bounds.

**Benefits:** This benefits those who have defined their geometries in a coordinate system different than the coordinates used for the data variable.

**Status Quo:**
According to Chapter 7, a data variable's coordinate variables can have a `nodes` attribute which identifies the equivalent variable from a geometry. Thus, nodes act as bounds to the coordinates, and bounds are considered as metadata to the coordinates. Therefore, following this logic, bounds can't have a different coordinate system.

**Detailed Proposal:**
Allow the different grid mapping on the geometry container variable, provided the auxiliary coordinates do not have the nodes as bounds, e.g. 

```
dimensions:
  instance = 2 ;
  node = 5 ;
  time = 4 ;
variables:
  int time(time) ;
    time:units = "days since 2000-01-01" ;
  double lat(instance) ;                     
    lat:units = "degrees_north" ;
    lat:standard_name = "latitude" ;
  double lon(instance) ;                  
    lon:units = "degrees_east" ;
    lon:standard_name = "longitude" ;
  int datum ;
    datum:grid_mapping_name = "latitude_longitude" ;
    datum:longitude_of_prime_meridian = 0.0 ;
    datum:semi_major_axis = 6378137.0 ;
    datum:inverse_flattening = 298.257223563 ;
  int Lambert_Conformal;
    Lambert_Conformal:grid_mapping_name = "lambert_conformal_conic";
    Lambert_Conformal:standard_parallel = 25.0;
    Lambert_Conformal:longitude_of_central_meridian = 265.0;
    Lambert_Conformal:latitude_of_projection_origin = 25.0;
  int geometry_container ;
    geometry_container:geometry_type = "line" ;
    geometry_container:node_count = "node_count" ;
    geometry_container:node_coordinates = "x y" ;
    geometry_container:grid_mapping = "Lambert_Conformal" ;
  int node_count(instance) ;
  double x(node) ;
[...]

Issue #155, opened 2019-03-01
Specify Geometry Dimension on Geometry Container


twhiteaker:
**Title:** Specify Geometry Dimension on Geometry Container

**Moderator:** @davidhassell 

**Requirement Summary:** 
This is in regard to geometries in Chapter 7.  For single part point geometries, the node_count variable is omitted since it is not needed. However, node_count is dimensioned using the instance dimension of the geometries, so without node_count, we don't know the geometry instance dimension.

Example of ambiguous geometry instance dimension (could be `foo` or `instance`):

```
dimensions:
    foo = 3 ;
    instance = 3 ;
    node = 3 ;
variables:
    int foo(foo);
        time:long_name = "bar" ;
    int geometry_container ;
        geometry_container:geometry_type = "point" ;
        geometry_container:node_coordinates = "x y z" ;
    double x(node) ;
        x:units = "degrees_east" ;
        x:standard_name = "longitude" ;
        x:axis = "X" ;
    double y(node) ;
        y:units = "degrees_north" ;
        y:standard_name = "latitude" ;
        y:axis = "Y" ;
    double z(node) ;
        z:units = "m" ;
        z:standard_name = "altitude" ;
        z:axis = "Z" ;
    double someData(instance, foo) ;
        someData:geometry = "geometry_container" ;
```

**Technical Proposal Summary:**
Add a `geometry_dimension` attribute to geometry container, which indicates the geometry dimension, as in:

```
dimensions:
    foo = 3 ;
    instance = 3 ;
    node = 3 ;
variables:
    int foo(foo);
        time:long_name = "bar" ;
    int geometry_container ;
        geometry_container:geometry_type = "point" ;
        geometry_container:node_coordinates = "x y z" ;
        geometry_container:geometry_dimension = "instance" ;
    double x(node) ;
        x:units = "degrees_east" ;
        x:standard_name = "longitude" ;
        x:axis = "X" ;
    double y(node) ;
        y:units = "degrees_north" ;
        y:standard_name = "latitude" ;
        y:axis = "Y" ;
    double z(node) ;
        z:units = "m" ;
        z:standard_name = "altitude" ;
        z:axis = "Z" ;
    double someData(instance, foo) ;
        someData:geometry = "geometry_container" ;
```

**Benefits:** Who or what will benefit from this proposal?  
Software developers can now properly parse the file; they know to which dimension the x, y, and z nodes apply.

**Status Quo:** Discussion of the current state CF and other standards.  
[...]

PR #158, opened 2019-03-04
geometry updates

twhiteaker:
Alternate grid mappings for geometry containers
https://github.com/cf-convention/cf-conventions/issues/156

Specify Geometry Dimension on Geometry Container
https://github.com/cf-convention/cf-conventions/issues/155


dblodgett-usgs:
This seems to all be agreed to and just waiting a merge so... 
843f040c437d38778aaa039c1609fe447383fd8796d5f273130d8a1cf61de7d4

PR #161, opened 2019-04-15
Formal improvements

zklaus:
Some minor improvements, fixing a bibliography issue and improving the pdf version of the document.

### DIFF
--- a/appf.adoc
+++ b/appf.adoc
@@ -622,7 +622,7 @@ to a reference direction.
 
 | **`straight_vertical_longitude_from_pole`** | N
 | The longitude (degrees_east) to be oriented straight up from the North or 
-    South Pole. Domain: **`-180.0 &le; straight_vertical_longitude_from_pole< 180.0`**
+    South Pole. Domain: **`-180.0 &lt;= straight_vertical_longitude_from_pole &lt; 180.0`**
 
 | **`towgs84`** | N
 |   This indicates a list of up 
--- a/bibliography.adoc
+++ b/bibliography.adoc
@@ -1,6 +1,6 @@
-[bibliography]
 == Bibliography
 
+[bibliography]
 === References
 
 - [[[COARDS]]]  link:$$http://www.ferret.noaa.gov/noaa_coop/coop_cdf_profile.html$$[ Conventions for the standardization of NetCDF Files ] .
--- a/cf-conventions.adoc
+++ b/cf-conventions.adoc
@@ -1,6 +1,8 @@
-﻿= NetCDF Climate and Forecast (CF) Metadata Conventions
-Brian Eaton; Jonathan Gregory; Bob Drach; Karl Taylor; Steve Hankin; Jon Blower; John Caron; Rich Signell; Phil Bentley; Greg Rappa; Heinke Höck; Alison Pamment; Martin Juckes; Martin Raspaud
+= NetCDF Climate and Forecast (CF) Metadata Conventions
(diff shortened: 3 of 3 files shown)


zklaus:
@painter1 could you have a look?

zklaus:
This should also fix the broken bibliography on the web page. Who is the right person to ask? @JonathanGregory, @japamment ?

martinjuckes:
@zklaus : is there a simple way to see the pdf document which will be generated after your updates? The improvements to the web document (correcting formatting errors in the bibliography) are good.

zklaus:
@martinjuckes thanks for following up! I attach the pdf here:
[cf-conventions.pdf](https://github.com/cf-convention/cf-conventions/files/3275585/cf-conventions.pdf).

It can be generated with

```
asciidoctor-pdf cf-conventions.adoc
```

martinjuckes:
@zklaus : thanks. I can see the improvement in Appendix F (page 122 of the revised document). When `asciidoctor-pdf` is run on the current conventions document there are 3 errors reported, with your modifications it runs cleanly.

On a style point: the use of `:media: prepress` forces sections to start on even pages, and this generates a lot of blank pages at the start of the document. I guess this is good if you want to turn it into a bound document, but I'm not sure it is ideal if it is just being used as an electronic document. If this line is left out, the same effect could be obtained with `asciidoctor-pdf --attribute media=prepress cf-conventions.adoc` .. so nothing is lost. How do you feel about leaving this line out?

zklaus:
I take your point about the empty pages and digging a bit deeper I found a different solution which results in the correct determination of right and left pages without the empty pages. The resulting document looks like [this](https://github.com/cf-convention/cf-conventions/files/3276394/cf-conventions.pdf). What do you think?

martinjuckes:
Thanks, that looks better to me.

If you know how to fix it, it would be nice to have the page numbering start after the Table of Contents, rather than in the middle. But this is a new issue, so don't waste time if there isn't an easy solution.

Otherwise, I support accepting this pull request.

zklaus:
I agree, unfortunately it seems not so easy. There is some development in the latest alpha release of asciidoctor-pdf, but I could not figure it out easily.

martinjuckes:
I looked into it a bit further: I haven't solved the numbering issue, but stumbled on another problem with the current conventions document which it might be worth fixing in your update. Some sites suggest converting to `docbook` format and then using `dblatex` to generate a pdf. If you try using `asciidoc`, a number of formatting error are flagged up and the program fails. `asciidoctor` will generate a `docbook` file, but `dblatex` considers it to be invalid. 

One of the errors reported by `asciidoc` concerns subsection 7.5 "geometries" which appears to be present in the `pdf` which you have generated, but absent from the online version of the conventions and the associated `pdf`. Although it is present, the presentation is not great because the length of the example is a challenging the format: I suspect this is an issue which can be postponed to another day. However, it would be worth making sure that section 7.5 gets into the online version. I think the error causing it to be omitted is the use of quotes in `[["timeseries-with-geometry"]]` (rather than `[[timeseries-with-geometry]]`) in `ch07.adoc`, in the header for example 7.15. Related to this, the indexing for examples 7.2 - 7.14 are wrongly numbered and examples 7.1 and 7.15 are missing from the index.

martinjuckes:
Or would you prefer to complete this update, and we can raise a new issue for the problems pre-existing problems I've described above?

zklaus:
I think it would be good to tackle those issues separately.

martinjuckes:
OK, I agree with that, and support accepting your changes as they are. I've raised an issue (#164) for the new problems mentioned above.

zklaus:
Great, thanks! Do I have to do something else now?

martinjuckes:
@painter1 , @japamment : can one of you pull this request into the master now?

dblodgett-usgs:
@zklaus -- once you merge the PR I opened to you, I'll go ahead and merge this.

PR #176, opened 2019-07-17
update to 1.8 and fix authors

dblodgett-usgs:
The header information here: is out of date. This fixes that.
http://cfconventions.org/cf-conventions/cf-conventions.html


dblodgett-usgs:
Going to close this and move updates to #161

Issue #144, opened 2018-08-03
Support of groups in CF


erget:
As discussed in CF meeting on 20 June 2018 in Reading, UK, we'd like to add support for the use of groups in CF files. Charlie and I have drafted an appropriate pull request containing the suggestions we'd like to implement.

Basically, the idea is to allow elements in CF files to refer to others which are not in the same group. The proposal defines ways of locating variables in this situation and tries to capture current ways of doing so "in the wild".

I'll update this issue with further material on this (Pull Request reference, summary presentation discussed at the meeting last June, and the proposal as drafted up till that point) as soon as the PR has an URL.

We haven't updated the conformance rules and checker yet, as we want to agree on the content first.

erget:
# Further resources:
- [The original proposal](https://github.com/diwg/cf2/blob/master/group/cf2-group.adoc) which was used to hash out ideas around these proposed changes
- [The presentation](https://github.com/cf-convention/cf-conventions/files/2257153/CF-2.Group.Proposal_.Advancing.netCDF-CF.for.the.Geosciences.pptx) discussed at the meeting in Reading

JonathanGregory:
This issue is implemented in https://github.com/cf-convention/cf-conventions/pull/145

JonathanGregory:
Dear Daniel and Charlie

Thanks for this proposal. I think it's a sensible way to support groups, and I agree with it, except that I'm not convinced about the lateral search. You don't sound convinced about it either, in saying that it's allowed for backward-compatibility and may be deprecated in future. Why not deprecate it *now* (meaning that the CF-checker would give a warning)? Why allow it at all? Is there a more specific extra search rule you could provide instead of your general lateral algorithm to deal with the existing datasets you have in mind?

I think that at the meeting we talked about how you would flatten a file, if you didn't want groups. Could we include that aspect in the proposal? If it's reversible, that would be neat.

I feel that the text could be made simpler and easier to understand, and most of my suggestions below are along those lines. Although I think I get the spirit of your search rules, I have to say I don't properly follow them in detail. The text seems obscure to me - maybe you could say it more plainly, and diagrams would help, I expect.

* In ch01 you propose a number of definitions. However, you don't propose to use these terms anywhere in the convention outside your new subsection of ch02. Hence I suggest you start your ch02 subsection with these definitions, instead of putting them in ch01, so they're in the place where the reader of ch02 will see them.

[...]

PR #180, opened 2019-07-17
Add authors of Groups proposal

erget:
See issue #144 for discussion of these changes. This PR is intended to work on top of #145 - it is a simple clerical update. @dblodgett-usgs as discussed.


dblodgett-usgs:
Looks like this will cause a conflict with #161 when it's merged. @zklaus, can you pull in these changes for your pull request too?

Issue #177, opened 2019-07-17
Automatic deployment isn't working.


dblodgett-usgs:
@dopplershift -- it looks like the automatic deployment has been broken for some time. I don't have access to the cf-bot. Can you make sense of how this is supposed to work?

marqh:
hello @dblodgett-usgs 

an initial look over 
https://github.com/cf-convention/cf-conventions/blob/master/.travis.yml
suggests that multiple failures have accreted over time.

the build is now working, e.g.
https://travis-ci.org/cf-convention/cf-conventions/builds/560188574
but I can't see where the build results are being sent.

I don't have access to `cf-conventions-bot`, afaik

As well as the ruby error which you fixed, there seems to be a disconnect regarding which repositories and branches are being targeted.

The cf website is being served by https://github.com/cf-convention/cf-convention.github.io, as far as I can tell.

However, the travis build never accesses this repository.
Instead, the travis build checks out the `gh-pages` branch from the cf-conventions project (I think)
https://github.com/cf-convention/cf-conventions/blob/master/.travis.yml#L24

It looks to me like it then tries to push the changes to the gh-pages branch of the cf-conventions repository:
https://github.com/cf-convention/cf-conventions/blob/master/.travis.yml#L41

This step succeeds, according to travis
https://travis-ci.org/cf-convention/cf-conventions/builds/560188574#L275

But, here's where the trail runs cold for me: 
https://github.com/cf-convention/cf-conventions/tree/gh-pages
hasn't been updated since August 2017
(co-incidentally automatically from the merger of my pull request https://github.com/cf-convention/cf-conventions/commit/679e3ed551f1b261b2d788a3e676b01c753143e6)
So, why has travis reported a successful push to gh-pages, but the repo hasn't picked it up.  I don't know.

It seems clear to me that the travis build is not communicating with the 
https://github.com/cf-convention/cf-convention.github.io
repository.  Perhaps this repo was created after the travis and cf bot were implemented.

However, it is far from clear to me how
http://cfconventions.org/latest.html
is serving 
http://cfconventions.org/cf-conventions/cf-conventions.html
If this is a github.io served website, then where in 
https://github.com/cf-convention/cf-convention.github.io
is cf-conventions/cf-conventions.html
I can't find it, and yet the URL:
http://cfconventions.org/cf-conventions/cf-conventions.html
resolves and provides a resource.
[...]

PR #182, opened 2019-07-18
Get automated build working again. And change title to "1.8 (draft)"

dblodgett-usgs:
See issue #177 for discussion of these changes. 

I've converted to use [travis environment variables](https://docs.travis-ci.com/user/environment-variables/) in the travis build settings. I've used a [personal github token](https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line) -- but anyone with push privileges to the repository could use their own in case I get hit by a truck. 

Screen Shot 2019-07-18 at 2 45 25 PM

I ended up keeping the configuration of the deployer the same as it was rather than [using the travis deployer.](https://docs.travis-ci.com/user/deployment/pages/). This does not use the @cf-conventions-bot at all. It just uses my personal access token with a made up name and email as in the screen shot above.

I tested it over on my personal repository. Pushed result can be seen here: https://dblodgett-usgs.github.io/cf-conventions/cf-conventions and here: https://dblodgett-usgs.github.io/cf-conventions/cf-conventions.pdf

Not going to merge this myself as I'd like at least one other user with admin privileges to have a look and make sure they know what I did! At this point that list is these people: https://github.com/orgs/cf-convention/people?utf8=%E2%9C%93&query=+role%3Aowner Maybe @ethanrd or @painter1 ??

PR #183, opened 2019-07-19
update bot token

marqh:
See issue #177 or discussion of these changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ after_success:
         git add $TARGET_DIR/index.html;
       fi
     - git commit -m "Auto-generated from ${TRAVIS_REPO_SLUG}@${TRAVIS_COMMIT}"
-    - git push https://${GH_TOKEN}@github.com/${TRAVIS_REPO_SLUG}.git gh-pages >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


marqh:
may be able to replace
#182 

however, this can only really be tested by merging it, as it will only run and use the token on master

if people are uncomfortable with this, I can make a branch on the main repository to prove the implementation, and make it push to another branch, but this is more work

marqh:
to keep the implementation clean, it would be good to remove the CI Environment variable changes made to facilitate
#182 as these are re-using environment variables, which could then conflict

I'm happy to do this before merge, if someone aims to merge this

consideration of this approach compared to #182 is useful, I feel

I hope this is helping

dblodgett-usgs:
I'll take care of it and go ahead and merge here. It's not going to get any more broken.

marqh:
```
 	cf-conventions.html 	Auto-generated from a884f3e 	14 seconds ago
	cf-conventions.pdf 	Auto-generated from a884f3e 	14 seconds ago
```

I believe we are cooking
:)

marqh:
thank you @dblodgett-usgs

Issue #136, opened 2018-07-17
missing trajectory dimension in h.22


castelao:
Example [H.22](https://github.com/cf-convention/cf-conventions/blob/master/apph.adoc) uses a dimension trajectory that is not declared.

Is that a real mistake? Shall I do a pull request adding trajectory to dimensions?

dblodgett-usgs:
Looks like a mistake to me! Want to PR the fix?

dblodgett-usgs:
Fixed in #185

davidhassell:
I agree that this is a defect, and the suggested fix is correct.

Sorry to be pain, but I believe that this shouldn't have been merged and closed yet, as the required three weeks' of "no objections" has not elapsed. As you know, I am uncomfortable with merging anything until we have an agreed strategy on how to do this. Verbal consensus appears to favour merging into the master branch once an issue is concluded, and the issue being closed (and I, too, am happy with this) - but this has not been formally decided upon and written up. This is perhaps the most pressing governance issue that needs resolving, as there are accepted pull requests that the community would benefit from being incorporated.

Thanks,
David

dblodgett-usgs:
@davidhassell -- my interpretation of the rules that we agreed to here: http://cfconventions.org/errors.html are that if there is no disagreement on an error in the document that we should feel free to modify the specification. Given that this issue lingered open with no objection for nearly a year, I saw it as a clear candidate to fall under this rule.  

That said, I think I've been in a move fast and break things mode -- a bit to push the community to actually _use_ github through a bit of learning so -- in this case, this may have warranted a bit more deliberate approach. Maybe lump this into the pile of things to consider in #172 and move on?

Best -- Dave

davidhassell:
Oh - I thought it was originally posted in 2019, not 2018! So 3 weeks is definitely up and it's perfectly OK to accept. Sorry, @castelao!

It would useful to say in the issue something like "This defect has gone disputed for > 3 weeks, so it is being accepted, I think", so that people can see that the rules are being followed,

I totally agree with your interpretation, @dblodgett-usgs, I'm just concerned that the procedure for "make the change and/or merge the pull request" has not been defined, yet. We'll carry this on off this issue...

Many thanks,
David

davidhassell:
3 weeks have passed with no further discussion, so we can accept this defect. Thanks, @castelao

PR #185, opened 2019-07-22
Missing dimension trajectory in H.22

castelao:
Variable trajectory uses dimension trajectory, which was missing.

See issue #136 for discussion of these changes.


dblodgett-usgs:
👍

Issue #198, opened 2019-08-15
Clarification of use of standard region names in "region" variables (Trac #151)


martinjuckes:
**Title:** Clarification of use of standard region names in "region" variables (Trac #151)
**Moderator:** @user  
**Requirement Summary:** 
The CF standard name region has the current description "A variable with the standard name of region contains strings which indicate geographical regions. These strings must be chosen from the standard region list." This description implies that the variable should be of character type, but it is often more convenient to have an integer variable and make a clear link to the region names using flag_values and flag_meanings. 
**Technical Proposal Summary:** 
The proposal is to clarify the definition so that either usage is acceptable and include an example of the latter usage in the convention text. It is also proposed that an appendix be added to the CF Convention text to state clearly any constraints on file meta-data which are implied by the CF Standard Name definitions, so that it is possible to test such constraints in the CF checker. 
**Benefits:** More flexibility in recording region information.
**Status Quo:** A natural approach to using flag values is ruled out on technical grounds. 
**Detailed Proposal:** See [Trac #151](https://cf-trac.llnl.gov/trac/ticket/151)
The issue was raised 3 years ago and agreement reached. It has not however, been implemented. This issue is provided to discuss a pull request which will be launched shortly.

Files to be modified are ch03.adoc and toc-extra.adoc (to update list of examples).

Updates to the descriptions of `region` and `area_type` in the standard name table will also be needed:

**region**

A variable with the `standard_name` of `region` contains either strings which indicate a geographical region or flags which can be translated to strings using `flag_values` and `flag_meanings` attributes. These strings are standardised. Values must be taken from the CF standard region list. 

**area_type**

A variable with the `standard_name` of `area_type` contains either strings which indicate the nature of the surface e.g. land, sea, sea_ice, or flags which can be translated to strings using `flag_values` and `flag_meanings` attributes. These strings are standardised. Values must be taken from the area_type table.

dblodgett-usgs:
@martinjuckes Unfortunately, Trac won't load for me right now. Can you identify a moderator for this issue to at least check that it is in line with what was agreed in Trac? Given that there has been no discussion for three weeks, I think we could consider this ready to merge as long as someone can verify it is in line with what was agreed in Trac.

taylor13:
@dblodgett-usgs -- Trac is back up again; sorry for the inconvenience.

dblodgett-usgs:
Thanks. I think this looks good. Will merge #199 and #200.

PR #200, opened 2019-08-15
Clarification of use of standard region names in "region" variables -…

martinjuckes:
…- new example

See #198 for discussion of these changes.


dblodgett-usgs:
Thanks for getting this material into a PR. I'm going to go ahead and merge so we can see it rendered in the draft document. @martinjuckes and others, do you feel this update warrants a note in the history?

martinjuckes:
@dblodgett-usgs : Thanks for merging this. I'm not sure of the approach for history notes .. but I think, for completeness, there should be a reference to the ticket 198: perhaps just the link to the ticket and its title.

PR #201, opened 2019-08-21
Updates to enhancement issue template

martinjuckes:
Updates to enhancement issue template in line with modified rules on moderation as discussed in cf-convention/cf-conventions#151

See issue #XXX for discussion of these changes.


dblodgett-usgs:
Looks fine to me. Will merge in three weeks if we don't get any further discussion.

PR #202, opened 2019-08-22
Fix Section 7 examples numbering in the list of examples

lbdreyer:
When the new example `Specifying formula_terms when a parametric coordinate variable has bounds` was added, the numbering in the list of examples was not updated.

I also noticed, in the same section, that Examples 7.15 and 7.16 were missing from the list of examples


dblodgett-usgs:
Thanks! It's been nearly three weeks with no comment and this is so clearly a defect fix, I'm going to go ahead and merge.

lbdreyer:
Thanks for merging @dblodgett-usgs !

Issue #203, opened 2019-08-29
Clarifications to use of groups


RosalynHatcher:
Following on from #144 which implemented groups, there are a couple of comments at the bottom of that issue which had been raised as a result of defining the conformance requirements for groups but have not yet implemented.  Summary of the comments and responses:

1. If paths are allowed in all areas of the convention it would be good to have a general statement that where "variable" or "dimension" is used this can either be a single variable name or a path to a variable. Perhaps we could also add path to the definitions in chapter 1? So that we don't have to define the term specifically in the conformance document.

2.  Although the CF standard doesn't standardize group names, which I understand is due to the abundance of NASA datasets that wouldn't meet this, I wonder whether it might be good to recommend that group names follow the same naming conventions as variables, dimensions and attributes (ie. begin with a letter and be composed of letters, digits and underscores). This way we would be encouraging human readable names but still allowing them not to be if required by some organizations.

Both of these were agreed in principle.

There was confirmation given that point 2 is indeed what NASA Dataset Interoperability Recommendations for Earth Science: Part 2 recommends.

RosalynHatcher:
I'll create a PR for review.

erget:
Hi @RosalynHatcher sorry about the long turnaround here - I'd seen this issue before and had noted down that I wanted to review the PR when it comes but haven't seen it yet.

I agree with these ideas still and would be willing to review and most likely endorse the PR :)

RosalynHatcher:
Hi @erget sorry about the delay on this I got diverted onto something else.  I'm going to try and have a CF day tomorrow and will pick this up then.

JonathanGregory:
@dblodgett-usgs wrote (in pull request https://github.com/cf-convention/cf-conventions/issues/203)
> With no dissent on this or #203 in a while, ok if I go ahead and merge? If no response by January 6th, I will.
Yes, please merge it. According to the rules, it's already well-qualified (in terms of time and support expressed)! Thanks for noticing.

Dave-Allured:
Here are two technical corrections for this issue.  Both are suggested changes to **2.3. Naming Conventions**.  I am not sure whether to start a new issue, but I thought I would start here for simplicity.  I believe these are both consistent with the intentions stated above.

1.  Change this sentence as follows.
`--- This convention does not standardize any variable or dimension names.`
`+++ This convention does not standardize any variable, dimension, or group names.`

[...]

PR #209, opened 2019-10-08
Clarification to use of groups

RosalynHatcher:
Variable and dimension names can be either a single variable name or a path to a variable
Add definition of path
Recommend that group names follow the same convention as variable and dimension names (ie. that they are composed of letters, digits and underscores)

See issue #203 for discussion of these changes.


dblodgett-usgs:
With no dissent on this or #203 in a while, ok if I go ahead and merge? If no response by January 6th, I will.

PR #224, opened 2020-01-06
coord value order for CRS WKT

### LINKED ISSUE #223: Axis Order for CRS-WKT grid mappings
**Title: Specify Axis Order for CRS-WKT grid mappings**

**Moderator:** @dblodgett-usgs   

**Moderator Status Review [last updated: 2020/01/05]:** 
A clear use case has emerged: If coordinate variable values and a CRS-WKT string are to be passed to a coordinate projection library, then the coordinate variables must be in the right order for the CRS-WKT string. Parsing the AXIS order of the CRS-WKT string and inferring how it relates to coordinate variable names is non-trivial if not impossible to automated in all cases. An explicit statement of coordinate variable axis order in a `grid_mapping` attribute of the form: `{variable}:grid_mapping = “{grid mapping}: {coordinate variable axis 1} {coordinate variable axis 2} ...”` should be required when using CRS-WKT. (@marqh’s [comment here](https://github.com/cf-convention/cf-conventions/issues/223#issuecomment-570593171) is a good place to look for more.)

I don’t see any strong disagreement to this being a valuable addition and think @marqh should go ahead and prepare a pull request. I would request that any modifications to the original proposal be applied to the description here at the top.

<details>
 <summary>Summary of comments.</summary>
A couple key comments below: 
@davidhassle asked: 
Are you saying the CRS-WKT stored as an attribute of the grid_mapping variable can contain the actual coordinate values in plain text, as well as the projection definition?

@marqh responded [more here](https://github.com/cf-convention/cf-conventions/issues/223#issuecomment-570189260):
The CRS definition does not contain any actual coordinate values. It is only the CRS definition.
The CRS-WKT definition includes a definition of a Coordinate System, a system of coordinates: which is a component part of a Coordinate Reference System. (careful with the similar but distinct terminology.

A minimal CDL example is:
```
dimensions:
  lat = 18 ;
  lon = 36 ;
variables:
  double lat(lat) ;
  double lon(lon) ;
  float temp(lat, lon) ;
    temp:grid_mapping = "crs:lat lon" ;
  int crs ;
    crs:grid_mapping_name = "latitude_longitude"
    crs:semi_major_axis = 6378137 ;
    crs:inverse_flattening = 298.257223563 ;
    crs:crs_wkt = "GEODCRS["WGS 84",
  DATUM["World Geodetic System 1984",
    ELLIPSOID["WGS 84",6378137,298.257223563,LENGTHUNIT["metre",1.0]]],
  CS[ellipsoidal,2],
    AXIS["latitude",north,ORDER[1]],
    AXIS["longitude",east,ORDER[2]],
    ANGLEUNIT["degree",0.01745329252], ID["EPSG",4326]]"
```
And @marqh stated:
This syntax is already available in CF, but it does not yet carry the meaning which I am suggesting that we add.

In a [useful response](https://github.com/cf-convention/cf-conventions/issues/223#issuecomment-570227603] @marqh stated: 
> All instances of CRS WKT string will be expected to have an AXIS list.
> 
> The intent of the extra interpretation is to ensure that a map between variables in the file and the concepts in the CRS WKT is explicit.

In the case of two grid mappings and two sets of (related) coordinate variables, the `grid_mapping` attribute would look like: `temp:grid_mapping = "crs:x y crs_wgs84:lat lon”`

It was noted that examples will be crucial here.
</details>

**Requirement Summary:** 

There is a requirement for using Coordinate Reference System Well Known Text to define coordinate reference systems that coordinate value sets be presented as ordered tuples, where each tuple represents a location with respect to that coordinate reference system.

Currently CF does not provide a mechanism to specify axis order when linking a data variable to a CRS-WKT instance via a grid mapping.

**Technical Proposal Summary:** 

**Benefits:** Who or what will benefit from this proposal?  

Users of CRS-WKT will benefit from this proposal as it will ensure that data sets comply with the OGC mandate to provide coordinate tuple value ordering as part of the reference between 

Whilst axis order can sometimes be inferred from deep inspection of the crs_wkt string and comparison to standard names of variables, this is a challenging inference to make and prone to significant problems, especially in automation scenarios.

I think that treating the WKT string as a parcel for parsing by other systems is much safer.

**Status Quo:** 

CF has adopted CRS-WKT from OGC but has not provided a mechanism to state how the variables containing coordinate values need to be structured in order to comply with the OGC mandates on implementation of Referencing by Coordinates in general, of which CRS-WKT is an implementation.

**Detailed Proposal:** 

Provide an explicit mechanism for defining axis order, to enable individual coordinate variables to be unambiguously expanded into coordinate tuple collections.

The expanded form of the grid_mapping attribute is 
> a blank-separated list of words "grid_mapping_variable: coordinate_variable [coordinate_variable …​] [grid_mapping_variable: …​]", which identifies one or more grid mapping variables, and with each grid mapping associates one or more coordinate_variables, i.e. coordinate variables or auxiliary coordinate variables.

This in situ feature provides a useful point to provide this capability extension.

As the extended form already explicitly references individual variables to use as coordinate values, the order of these values may be used to explicitly state the axis order of the coordinate values defined with respect to the coordinate reference system.

Explicitly, the following text may be added as following paragraphs.

> Where an extended "grid_mapping_variable: coordinate_variable [coordinate_variable …​]" entity is defined, then the order of the coordinate variable definitions provides an explicit order for these coordinate value variables to be combined into coordinate tuples.  

> The order of the coordinate_variable instances defines the order of elements within a coordinate value tuple.  This enable an application reading the data from a file to construct an array of coordinate value tuples, where each tuple is ordered to match the specification of the coordinate reference system being used.

> This explicit 'axis order' is important when the `grid_mapping_variable` contains an attribute `crs_wkt`.  It is mandated by the OGC CRS-WKT standard that coordinate tuples with correct axis order are provided as part of the reference to a Coordinate Reference System.

I would also suggest including text within the later `Use of the CRS Well-known Text Format` section:

> Where `crs_wkt` is added to a grid_mapping, the extended syntax for the `grid_mapping` attribute enables the axis order for the coordinates being referenced to be explicitly stated.  The explicit definition of axis order is expected by the OGC standards for referencing by coordinates.

marqh:
See issue #223  for discussion of these changes.

Fixes #223 

Provides a mechanism to specify explicit 'Axis Order' for use with CRS WKT using enhanced interpretation of existing syntax

### DIFF
--- a/ch05.adoc
+++ b/ch05.adoc
@@ -207,10 +207,15 @@ The grid mapping variables are associated with the data and coordinate variables
 **`grid_mapping`** attribute. This attribute is attached to data variables so that variables with
 different mappings may be present in a single file. The attribute takes a string value with two
 possible formats. In the first format, it is a single word, which names a grid mapping variable. In
-the second format, it is a blank-separated list of words "grid_mapping_variable: coordinate_variable
-[coordinate_variable ...] [grid_mapping_variable: ...]", which identifies one or more grid mapping
-variables, and with each grid mapping associates one or more coordinate_variables, i.e. coordinate
+the second format, it is a blank-separated list of words "grid_mapping_variable: coordinates_variable
+[coordinates_variable ...] [grid_mapping_variable: ...]", which identifies one or more grid mapping
+variables, and with each grid mapping associates one or more coordinates_variables, i.e. coordinate
 variables or auxiliary coordinate variables.
+
+Where an extended "grid_mapping_variable: coordinates_variable [coordinates_variable]" entity is defined, then the order of the coordinates_variable references within the definition provides an explicit order for these coordinate value variables, used if they are to be combined into individual coordinate tuples.
+
+This order is only significant if crs_wkt is also specified within the referenced grid mapping variable. Explicit 'axis order' is important when the grid_mapping_variable contains an attribute crs_wkt as it is mandated by the OGC CRS-WKT standard that coordinate tuples with correct axis order are provided as part of the reference to a Coordinate Reference System.
+
 
 Using the simple form, where the **`grid_mapping`** attribute is only the name of a grid mapping
 variable, 2D latitude and longitude coordinates for a projected coordinate reference system use the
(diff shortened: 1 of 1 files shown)


marqh:
Note: I have examined the conformance document and I propose no updates to that document as a result of this change.

Enforcing conformance is too detailed a process, likely involving parsing of the CRS WKT string, which is not in scope for conformance checking

marqh:
many thanks @davidhassell

Issue #212, opened 2019-11-07
Inconsistent usage of false_easting and false_northing in grid mappings definitions and in examples


neumannd:
# Summary
## Title
Inconsistent usage of `false_easting` and `false_northing` in grid mappings definitions and in examples
## Moderator
@erget 
## Moderator status review
Explicit support from
- @JimBiardCics 
- @marqh 
- @JonathanGregory 
- @erget 

No objections expressed in this discussion.

https://github.com/SciTools/iris/issues/3626 notes that the current inconsistency sometimes leads to problems; @erget has proposed that solving the ambiguity would at least make the intent of the Conventions clear. Impact: Software would need to be aware of the defaults.

## Requirement Summary
Definitions of grid mappings / projections and their usage in examples should be consistent with respect to required attributes. 

## Technical Proposal Summary
As in #225.

Add `false_easting` and `false_northing` to [example 5.7](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#lambert-conformal-projection) or make them optional in the definition of all projections where applicable (e.g. [Lambert Conformal Conic projection](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_lambert_conformal)).

## Benefits
Consistency 
## Status Quo
The attributes `false_easting` and `false_northing` are required in the grid mapping variable for a [Lambert Conformal Conic projection](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_lambert_conformal). [Example 5.7](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#lambert-conformal-projection) does not use `false_easting` and `false_northing`.
## Detailed Proposal
Either `false_easting` and `false_northing` should be made optional in the definition of the projections (assumed to be zero if not set) or they should be added to example 5.7. I would prefer the first option as it is already the case for the attribute `north_pole_grid_longitude` in the grid mapping for [rotated poles](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_rotated_pole):

> `north_pole_grid_longitude` - This parameter is option (default is 0).

Hence, I would suggest to replace 

> `false_easting`
> `false_northing`

in **all** grid mapping definitions by

> `false_easting` - This parameter is option (default is 0).
> `false_northing` - This parameter is option (default is 0).

neumannd:
[...]

PR #225, opened 2020-01-07
Update usage of false_easting and false_northing in projections

neumannd:
The projection parameters `false_easting` and `false_northing` should be considered as optional for all projections. If they do not exist, they should be assumed to be `0`. This was added to the definitions of these parameters.

See issue #212 for discussion of these changes.

### DIFF
--- a/appf.adoc
+++ b/appf.adoc
@@ -433,9 +433,8 @@ to a reference direction.
 |  Applied to all abscissa values in the rectangular
     coordinates for a map projection in order to eliminate negative numbers. Expressed in
     the unit of the coordinate variable identified by the
-    standard name **`projection_x_coordinate`**. If 
-    **`false_easting`** is not provided it is assumed 
-    to be 0.
+    standard name **`projection_x_coordinate`**. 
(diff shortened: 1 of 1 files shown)


appf.adoc
```diff
@@ -433,7 +433,9 @@ to a reference direction.
 |  Applied to all abscissa values in the rectangular
     coordinates for a map projection in order to eliminate negative numbers. Expressed in
     the unit of the coordinate variable identified by the
-    standard name **`projection_x_coordinate`**.
+    standard name **`projection_x_coordinate`**. If
```
erget: @neumannd could you move this sentence onto its own line (and on line 449) so it's one sentence, one line?

In source code it makes sense to restrict line length, but for a document like the Conventions it makes it easier to compare changes if each sentence stands on its own line. Then a blame doesn't produce false positives. We don't need to reflow all the text in the documents but when we touch or add a sentence I think it's best to put them on their own line.

neumannd:
@erget: Thanks for the info. I updated the text accordingly.

Issue #228, opened 2020-01-14
Sections in issue templates


erget:
When I make issues following the issue template I find myself hampered by the fact that our templates use formatting, not sections, to construct the form. I'd like to use sections so that paragraph breaks are more obvious and have implemented this in the "Defect" and "Enhancement" templates in the attached PR. Any objections?

davidhassell:
No objections from me.

ethanrd:
# No strong objections
My only hesitation is that it will take up more vertical space on the page. I tend to like this type of summary text to be pretty tight vertically to minimize the need for scrolling.

**Formatting vs Sections:** I had to play around a bit with formatting and sections to figure out what you meant.

Also played around with
single returns which don't really give a clear paragraph break.

And with blank lines which do give a pretty good paragraph break.

**Summary:** I guess I would prefer formatting and blank lines for clear paragraph breaks.

However, the blank line for paragraphs would be hard to enforce whereas using sections makes it a bit more cut and dry.

**Summary 2:** So, I'm OK either way.

erget:
**Alternative**
We could also just move the formatted text to its own line and save space.

**Pro**
You save some vertical space and still have paragraph breaks.

**Con**
It doesn't allow making subsections of e.g. summaries. But one could still use formatting. And this isn't such a common thing to do.

Since nobody seems to have a strong opinion specifically about headers, I'll modify the PR accordingly.

erget:
...Waffling here, as I do find sections more cut and dry. @ethanrd I'm going with your first summary and leaving it for the moment ;) I'll wait with the merge until Friday though so anyone else can give their thoughts.

PR #229, opened 2020-01-14
Use sections not formatting to allow paragraphs

erget:
See issue #228 for discussion of these changes.


dblodgett-usgs:
Fine by me. There was no reason to use **strong** text rather than titles when those were created AFAIK.

JimBiardCics:
Sounds great to me.

PR #232, opened 2020-01-17
Fix geostationary projection

erget:
See issue #230 for discussion of these changes.

### DIFF
--- a/appf.adoc
+++ b/appf.adoc
@@ -79,7 +79,8 @@ __Map parameters:__::
 * **`sweep_angle_axis`**
 * **`fixed_angle_axis`**
 
-__Map coordinates:__:: The x (abscissa) and y (ordinate) rectangular coordinates are identified by the **`standard_name`** attribute values **`projection_x_angular_coordinate`** or **`projection_x_coordinate`** (deprecated) and **`projection_y_angular_coordinate`** or **`projection_x_coordinate`** (deprecated), respectively.
+__Map coordinates:__:: The x (abscissa) and y (ordinate) rectangular coordinates are identified by the **`standard_name`** attribute values **`projection_x_angular_coordinate`** and **`projection_y_angular_coordinate`**, respectively.
+CF specified the standard names **`projection_x_coordinate`** and **`projection_y_coordinate`** for these coordinates prior to version 1.9, but that use is deprecated.
 In the case of this projection, the projection coordinates in this projection are directly related to the scanning angle of the satellite instrument.
 
 __Notes:__:: The algorithm for computing the mapping may be found at link:$$http://www.cgms-info.org/documents/pdf_cgms_03.pdf$$[http://www.cgms-info.org/documents/pdf_cgms_03.pdf].
--- a/appf.adoc
+++ b/appf.adoc
@@ -101,7 +101,7 @@ The **`fixed_angle_axis`** attribute indicates which axis the instrument is fixe
 The values are opposite to **`sweep_angle_axis`**. Only one of those two attributes is mandatory.
 
 +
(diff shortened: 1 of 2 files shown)


appf.adoc
```diff
@@ -79,17 +79,30 @@ __Map parameters:__::
 * **`sweep_angle_axis`**
 * **`fixed_angle_axis`**
 
-__Map coordinates:__:: The x (abscissa) and y (ordinate) rectangular coordinates are identified by the **`standard_name`** attribute values **`projection_x_coordinate`** and **`projection_y_coordinate`** respectively. In the case of this projection, the projection coordinates in this projection are directly related to the scanning angle of the satellite instrument, and their units are radians.
+__Map coordinates:__:: The x (abscissa) and y (ordinate) rectangular coordinates are identified by the **`standard_name`** attribute values **`projection_x_angular_coordinate`** or **`projection_x_coordinate`** (deprecated) and **`projection_y_angular_coordinate`** or **`projection_x_coordinate`** (deprecated), respectively.
+In the case of this projection, the projection coordinates in this projection are directly related to the scanning angle of the satellite instrument.
```
lesserwhirls: Perhaps the intent of the proposal is clouded by the removal of the explicit use of radians in this sentence, although it is still called out explicitly in line 104 as part of the explanation of the change in starting in 1.9? If so, would it help to put it back so it's right up front?

erget: I feel torn here because the units are actually up to the user, whereas they must be compatible with the canonical units. Those are connected to the `standard_name` used and in order to reduce redundancy I wouldn't call it up here.

I think they were mentioned before because the paragraph was essentially overwriting the standard names table's values in this special case.

Optimally, from 1.9 I'd like to completely remove the use of the old units rather than simply deprecating it but that's probably a bridge too far.

appf.adoc
```diff
@@ -79,17 +79,30 @@ __Map parameters:__::
 * **`sweep_angle_axis`**
 * **`fixed_angle_axis`**
 
-__Map coordinates:__:: The x (abscissa) and y (ordinate) rectangular coordinates are identified by the **`standard_name`** attribute values **`projection_x_coordinate`** and **`projection_y_coordinate`** respectively. In the case of this projection, the projection coordinates in this projection are directly related to the scanning angle of the satellite instrument, and their units are radians.
+__Map coordinates:__:: The x (abscissa) and y (ordinate) rectangular coordinates are identified by the **`standard_name`** attribute values **`projection_x_angular_coordinate`** or **`projection_x_coordinate`** (deprecated) and **`projection_y_angular_coordinate`** or **`projection_x_coordinate`** (deprecated), respectively.
```
davidhassell: This sentence has a lot of "or"s and "and"s. Perhaps an extra comma after the first "(deprecated)" might help parsing?

__Map coordinates:__:: The x (abscissa) and y (ordinate) rectangular coordinates are identified by the **`standard_name`** attribute values **`projection_x_angular_coordinate`** or **`projection_x_coordinate`** (deprecated), **[[new comma here]]** and **`projection_y_angular_coordinate`** or **`projection_x_coordinate`** (deprecated), respectively.

JimBiardCics: How about a more direct approach? I think it will be less confusing.

__Map coordinates:__:: The x (abscissa) and y (ordinate) rectangular coordinates are identified by the **`standard_name`** attribute values **`projection_x_angular_coordinate`** and **`projection_y_angular_coordinate`** respectively. CF specified the standard names **`projection_x_coordinate`** and  **`projection_y_coordinate`** for these coordinates prior to version 1.9, but that use is deprecated.

davidhassell: That's much better! Thanks

erget: Implemented in c73544e

PR #236, opened 2020-01-23
Amend broken link (RE COARDS Conventions) in bibliography

sadielbartholomew:
No corresponding issue, as this is a uncontroversial one-line change to fix a broken link.

See also Issue cf-convention/cf-convention.github.io#74 which registers an equivalent reference to the obsolete URL on the website, & where I have commented regarding the choice of replacement link (in short, with a change of the path as well as for the new NOAA Ferret host, the same content is kept included but high-up on the core sitemap).

### DIFF
--- a/bibliography.adoc
+++ b/bibliography.adoc
@@ -3,7 +3,7 @@
 [bibliography]
 === References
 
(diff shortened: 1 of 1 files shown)


dblodgett-usgs:
🙏 and sorry for the delay in merging.

PR #251, opened 2020-03-18
Check that PRs do not break Asciidoctor build

lesserwhirls:
This PR enables a github action to check that PRs do not break the asciidoc build. If the build is successful, the pdf and html files are uploaded as artifacts to the action, which allow for easier previews of the generated files.

### DIFF
--- a/.github/workflows/check_adoc_build.yml
+++ b/.github/workflows/check_adoc_build.yml
@@ -1,0 +1,68 @@
+#
+# GitHub Actions Workflow for pull request events against the master branch to
+# make sure Asciidoctor still builds. If successful, the resulting html and
+# pdf files are uploaded as artifacts to make it easier to preview the results
+# of a PR.
+#
+# For more information on the actions used in this workflow, please see:
+#   https://github.com/actions/checkout 
+#   https://github.com/Analog-inc/asciidoctor-action
+#   https://github.com/actions/upload-artifact
+#
+name: Check Asciidoctor Build
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  # Job to build cf-conventions.html, cf-conventions.pdf
+  build_conventions:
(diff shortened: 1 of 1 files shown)


lesserwhirls:
As an example of what this will look like, see [here](https://github.com/lesserwhirls/cf-conventions/actions/runs/58486955).

marqh:
I agree this is a useful and sensible step

thank you for the input @lesserwhirls

lesserwhirls:
Thanks @marqh! Happy to help where I can :-)

zklaus:
This is great!
Would it be possible to upload the pdf and html individually, not in an archive so as to facilitate easy inspection?
Maybe it would also be a good idea to upload a log (perhaps just the console output)?
Like this, we might catch some warnings more easily that don't prevent a successful creation, but would be nice to know anyways.

lesserwhirls:
> This is great!
> Would it be possible to upload the pdf and html individually, not in an archive so as to facilitate easy inspection?

That would be great! Currently, the `actions/upload-artifact` github action does not support it, but it's on their list of improvements (more info [here](https://github.com/actions/upload-artifact/issues/3#issuecomment-598820814)). 

> Maybe it would also be a good idea to upload a log (perhaps just the console output)?
> Like this, we might catch some warnings more easily that don't prevent a successful creation, but would be nice to know anyways.

I've updated this PR to enable the asciidoctor build to use the `--verbose` flag, which seems to show warnings. An example of what the output looks like now can be found [here](https://github.com/lesserwhirls/cf-conventions/runs/528953848#step:5:5).

ethanrd:
Two positive reviews and it looks good to me. I'm going to merge.

I will look at how to require reviews for this type of PR, probably by members of the [info-mgmt](https://github.com/orgs/cf-convention/teams/info-mgmt/members) team - @cf-convention/info-mgmt

PR #268, opened 2020-05-13
Removing unnecessary dimensions

cofinoa:
In some examples where only `string` type is been used, there is a dimension `name_strlen` which it's unnecesary.


JonathanGregory:
Thanks for noticing and correcting this, @cofinoa. It looks OK to me. Please could you open a defect issue for it as well, linking this pull request?

cofinoa:
@JonathanGregory I have created the issue #272

PR #270, opened 2020-05-21
Extend the data model for Geometries

davidhassell:
See issue #271 for discussion of these changes.


davidhassell:
See issue #271 for details

appi.adoc
```diff
@@ -290,49 +290,101 @@ corresponding to a different field construct.
...
-variables correspond to dimension coordinate constructs.
+locate the cells of the domain and which depend on a subset of the
+domain axis constructs. A coordinate construct consists of an optional
+data array of the coordinate values spanning the subset of the domain
+axis constructs, properties to describe the coordinates (in the same
+sense as for the field construct), an optional data array of cell
+bounds recording the extents of each cell, and any extra arrays needed
+to interpret the cell bounds values. In CF-netCDF, the data array of
+the coordinate values is required, execpt for the special cases
+described below.
```
davidhassell, line 301: The "In CF-netCDF, " text should not be there, as it could conceivably imply that the array of coordinate values is always optional in the data model, which it is not.

appi.adoc
```diff
@@ -290,49 +290,100 @@ corresponding to a different field construct.
...
-coordinate values. These axes are called "discrete" axes in CF-netCDF.
+axis that indexes a time series at scattered points. These axes are
+discrete axes in CF-netCDF. In such cases cells may be described with
+one-dimensional auxiliary coordinate constructs for which, provided
+that there is a cell bounds array to describe the cell extents, the
+coordinate array is optional, since coordinates are not always well
+defined for such cells. A CF-netCDF geometry container variable is
+used to store cell bounds without coordinates for a discrete axis.
+
+In CF-netCDF, when a CF-netCDF geometry container variable is present,
```
dblodgett-usgs, line 369: Redundant CF-NetCDF here.

appi.adoc
```diff
@@ -290,49 +290,100 @@ corresponding to a different field construct.
...
+it explicitly describes the geometry type and the number of parts per
+cell for the coordinate variables that it references. In this case, a
```
dblodgett-usgs, line 371: The geometry container variable indicates which variables contain the number of parts per cell -- it doesn't do it its self.

appi.adoc
davidhassell: Fixed

appi.adoc
davidhassell: Good catch, thanks. Text updated.

PR #282, opened 2020-07-06
Allow CRS WKT to represent the CRS without requiring reader to compare with grid mapping parameters

### LINKED ISSUE #222: Allow CRS WKT to represent the CRS without requiring reader to compare with grid mapping parameters
**Title:** Allow CRS WKT to represent the CRS without requiring reader to compare with grid mapping parameters
**Moderator:** ???  
**Moderator Status Review [last updated: YY/MM/DD]:** ???
**Requirement Summary:**

I propose the requirement be changed like so:

There will be occasions when a given CRS property value is duplicated in both a single-property grid mapping attribute and the crs_wkt attribute. In such cases the onus is on data producers to ensure that the property values are consistent. *If both crs_wkt and grid mapping attributes exist, the attributes must be the same and grid mapping parameters should always be completed as fully as possible. As such, information from either one (or both) may be read in by the user without needing to check both. However, in those situations where the two values of a given property are different, the CRS information cannot be interpreted accurately and users should inform the provider so the issue can be addressed.* ~~, then the value specified by the single-property attribute shall take precedence.~~ For example, if the semi-major axis length of the ellipsoid is defined by the grid mapping attribute semi_major_axis and also by the crs_wkt attribute (via the WKT SPHEROID[…​] element), *the value of this attribute cannot be interpreted accurately.*  ~~then the former, being the more specific attribute, takes precedence.~~ Naturally if the two values are equal then no ambiguity arises.

**Benefits:**
1. The CRS could originate from several different formats such as WKT, PROJ, or SRS Authority Code. If there are errors in the conversion process to the CF or WKT representation, only the provider would have the original CRS representation. As such, if there are conflicts, the provider would be the best source to go to in order to resolve the conflicts.
2. Making this change will simplify the lives of software developers so they can just read in the WKT or grid mapping CF parameters for the CRS without a need to compare the two.

**Status Quo:**
http://cfconventions.org/cf-conventions/cf-conventions.html#use-of-the-crs-well-known-text-format mentions

```    
There will be occasions when a given CRS property value is duplicated in both a single-property grid mapping attribute and the crs_wkt attribute. In such cases the onus is on data producers to ensure that the property values are consistent. However, in those situations where two values of a given property are different, then the value specified by the single-property attribute shall take precedence. For example, if the semi-major axis length of the ellipsoid is defined by the grid mapping attribute semi_major_axis and also by the crs_wkt attribute (via the WKT SPHEROID[…​] element) then the former, being the more specific attribute, takes precedence. Naturally if the two values are equal then no ambiguity arises.
```

snowman2:
The three weeks have gone by without complaint, so I am moving forward with this change.

# Release checklist
- [x] Closes #222 (See this issue for discussion of these changes)
- [ ] Next version in `cf-conventions.adoc` up to date? Versioning inspired by [SemVer](https://semver.org).
- [ ] Conformance document up-to-date?

# For maintainers

### DIFF
--- a/cf-conventions.adoc
+++ b/cf-conventions.adoc
@@ -37,6 +37,7 @@ include::toc-extra.adoc[]
 * Charlie Zender, University of California, Irvine
 * Daniel Lee, EUMETSAT
 * David Hassell, NCAS and University of Reading
+* Alan D. Snow, Corteva Agriscience
 
 Many others have contributed to the development of CF through their participation in discussions about proposed changes.
 
--- a/ch05.adoc
+++ b/ch05.adoc
@@ -477,23 +477,20 @@ For example, a file has two coordinate variables, lon and lat, and a grid mappin
 
 The **`crs_wkt`** attribute is intended to act as a _supplement_ to other single-property CF grid
 mapping attributes (as described in Appendix F); it is not intended to replace those attributes. If
-data producers omit the single-property grid mapping attributes in favour of the compound
-**`crs_wkt`** attribute, software which cannot interpret **`crs_wkt`** will be unable to use the
-grid_mapping information. Therefore the CRS should be described as thoroughly as possible with the
-single-property attributes as well as by **`crs_wkt`**.
+data producers omit the single-property grid mapping attributes in favour of the **`crs_wkt`**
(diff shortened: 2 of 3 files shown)


JonathanGregory:
https://github.com/cf-convention/cf-conventions/issues/222This proposal has been accepted in

Issue #301, opened 2020-09-22
Introducing a CF domain variable


davidhassell:
# Introducing a CF domain variable

# Moderator
@dblodgett-usgs 

# Moderator Status Review [last updated: 2020-10-15]
The proposal has been submitted and preliminarily reviewed by the moderator. Attention should be called to the potential for this proposal to subtly but fundamentally alter how CF-NetCDF data fields and domains are treated. Review from authors of CF-NetCDF client software is necessary here.

1. By in large, the community is supportive of the proposal.
1. There has been discussion of how to identify a `domain` variable: by `cf_role: domain` or by presence of a `dimensions: "X Y Z ..."` attribute. Presence of a `dimensions` attribute has won out for it's lack of redundancy.
1. The title of section 5 will now be: `"Coordinate systems and domain"`
1. There is some nuanced discussion of domain constructs for scalar (single-valued dimensionless / degenerate) coordinate variables. No major issues have been noted.
1. There is discussion of multiple domain variables for a single domain. No major issues have been noted.
1. The idea of adding a `domain: domain_variable` attribute on a data variable was suggested. Adding it would introduce redundancy and seems to be the wrong path.

As of 10-15-2020, discussion is slow but ongoing. I will check back in around the beginning of November.

# Requirement Summary

The concept of a domain that describes data locations and cell properties is not currently mentioned in the CF conventions, because it does not correspond to any single entity in the netCDF file. Instead, the domain is stored implicitly in a number of other variables and attributes that are linked to the data variable in various ways defined by the conventions.

The domain is, however, well defined in the CF data model as an abstract concept (as opposed to a data model construct) that provides the linkage between the field construct and the metadata constructs that describe the relevant data locations and cell properties. There is currently no "domain construct" in the data model, since there is no corresponding CF-netCDF entity.

There is a need to be able to describe a domain independently of any data variables, which is currently not possible. Use cases include:

* Curated data streaming services for which it is impractical to send very large domain descriptions with every file.

* Storing time-dependent coordinates from remote sensing applications.

* Storing geometries without any timeseries data.

For such use cases, it is not satisfactory to try to locate an appropriate multidimensional data variable that describes the required domain, nor to create a dummy data variable for this purpose, which has no physical meaning.

Therefore, the inclusion of CF-netCDF domain variables that can encode a domain independently of any data, and a corresponding data model domain construct, will enhance CF by meeting these use cases.

# Technical Proposal Summary

## NetCDF encoding

[...]

PR #302, opened 2020-09-22
Introducing a CF domain variable

davidhassell:
See issue #301 for discussion of these changes.

# Release checklist
- [x] Next version in `cf-conventions.adoc` up to date? Versioning inspired by [SemVer](https://semver.org).
- [x] Conformance document up-to-date?

# For maintainers


ch05.adoc
```diff
@@ -747,3 +747,291 @@ data:
...
+applications that have no need of data values at the domain's
+locations, thus removing any ambiguity when retrieving a domain from a
+dataset. Ancillary variables and cell methods are not part of the
+domain, because they are only defined in relation to data values.
+
+The domain variable supports the same attributes as are allowed on a
+data variable for describing a domain, with exactly the same meanings
+and syntaxes, as described in <<attribute-appendix>>. If an attribute
+is needed by a particular data variable to describe its domain, then
+that attribute would also be needed by the equivalent domain variable.
```
dblodgett-usgs, line 772: Having domain variables adopt from field variables in this way may become problematic. I think this is probably the right -- pragmatic -- approach for now, but we need to be careful with sweeping statements like is made here.

ch05.adoc
```diff
@@ -747,3 +747,291 @@ data:
...
+attribute. It is necessary to list these dimensions, rather than
+inferring them from the contents of the other attributes, as it can
+not be guaranteed that the referenced variables span all of the
+required dimensions (as could be the case for a discrete axis, for
+instance). The value of the **`dimensions`** attribute is a blank
+separated list of the dimensions names. There is no restriction on the
+order in which the dimensions appear in the **`dimensions`** attribute
+string. The presence of a **`dimensions`** attribute will identify the
+variable as a domain variable. If a domain has no dimensions then the
+value of the **`dimensions`** attribute must be an empty string.
```
dblodgett-usgs, line 784: Can a domain exist without any dimensions?

davidhassell: (I should have read the existing text first rather than replying from memory - sorry restated a lot of stuff uncontested stuff. I'll try again)

A domain can certainly exist without _explicit_ dimensions, as they may be implied by scalar coordinate variables.

Would this be valid domain, though?
```
    char domain ;
        domain:dimensions = "" ;
        domain:grid_mapping = "lambert_conformal" ;
```
My gut feeling is "yes" ....

ch05.adoc
davidhassell: I appreciate your concern, here. I feel that the approach is guaranteed to work for as long as CF still insists that a data variable encodes is own, implicit domain, which is to say for the foreseeable future. 

My one point of concern, however, is for a scalar data variable that innocently has the `dimensions` attribute - how can we tell if it is a data variable or a domain variable?

Perhaps using the `dimensions` attribute as the identifying feature could be replaced by the existing `cf_role` attribute. I think that this could be a bit safer:
``` 
 char domain ;
        domain:cf_role = "domain" ;
        domain:dimensions = "x y" ;
        domain:coordinates = "lat lon" ;

dblodgett-usgs:
If no further discussion takes place in #301, we will merge this pull request on Monday November 23rd.

ch05.adoc
erget: Resolved in https://github.com/cf-convention/cf-conventions/issues/301#issuecomment-702695406

erget (approved):

JonathanGregory (changes requested):
I think the content is fine, thanks, but I would note two minor problems in wording:

  * "blank separated list of the dimensions names" should be _dimension_ or _dimensions'_

  * In "Multiple domain variables may exist in a file with, or without, other data variables" I think "other" should be deleted, because domain variables are not data variables.

Jonathan

dblodgett-usgs (approved):

dblodgett-usgs:
@davidhassell -- please update the history then we can go ahead and merge IMHO.

davidhassell:
Thanks, @dblodgett-usgs. History is updated.

PR #310, opened 2020-12-14
Suggested reqording  of zones/areas 

davidhassell:
Hi Anders, some thoughts on zones. 

I dropped the line about how you might find discontinuites (either by inspection or knowledge of the sensor) only because I couldn't find a way to make it work, given that it doesn't affect the definitions being discussed. Is it perhaps more of an implementation detail, and so might be more at home in the appendix?

Thanks, David


davidhassell:
Submitted by mistake - sorry!

Issue #343, opened 2022-01-02
Single-source the version number


adigitoleo:
The current published standard (v1.9) contains a typo in section 1.4 Overview:

> Files using this version of the CF Conventions must set the NUG defined attribute Conventions to contain the string value "CF-1.8" to identify datasets that conform to these conventions.

So, the version number was not incremented in that paragraph. This could be avoided by somehow reading the version number from a central location when building the document, if that is possible.

Just something I noticed when reading through again today.

6 Jan 2022. I'm adding the enhancement label, because we are considering https://github.com/cf-convention/cf-conventions/pull/344 as a solution, and that's a substantive change. @JonathanGregory

zklaus:
This should be possible using a "replacement" as described in [Section 26.9 of the asciidoc userguide](https://asciidoc-py.github.io/userguide.html#X7).

erget:
Hi all, so it looks like we actually have 2 things in this issue:

1. Fix typo (easy)
2. Potentially improve the document so it's more future-proof

(1) sounds necessary to me; (2) sounds like a good idea. Is anybody willing to champion 1 or both of these? I would support it!

JonathanGregory:
Dear all

Thanks for pointing it out, Leon @adigitoleo. I think @zklaus is right that we could add a "replacement" i.e. a sort of macro in AsciiDoc for the current version number, so that only one occurrence needed to be changed. I am not familiar with the build process so I won't volunteer to do that. Is there a checklist for making a new version, to which this could be added? I think @davidhassell must know.  As Daniel @erget says, it would be easy to fix the typo, and I'm sure we would approve the change as a defect issue for it . But could we agree to modify the live 1.9 document _immediately_? There would be no point in agreeing a defect correction to 1.9 which wasn't going live until 1.10 in this case!

Best wishes and happy 2022 to the CF community

Jonathan

zklaus:
I added a draft PR in #344 that addresses (2). If we decide to tackle (1) separately, the list of changes in that PR should give a good overview of the places that need changing.

erget:
Thanks @zklaus, the PR looks pretty cool.

Also yes, happy new year everybody! :)

I've requested also the input of @davidhassell since he's more heavily involved in the release process. @JonathanGregory, I propose that we proceed as follows:
- If we can agree on an approach for a longer-term fix, e.g. by accepting the changes proposed in #344, then we cherry-pick those on top of 1.9 and make a patch release to correct the defect.
[...]

PR #344, opened 2022-01-04
Single source version

zklaus:
See issue #343 for discussion of these changes.

# Release checklist
- [x] Next version in `cf-conventions.adoc` up to date? Versioning inspired by [SemVer](https://semver.org).
- [x] Conformance document up-to-date?

# For maintainers

### DIFF
--- a/version.adoc
+++ b/version.adoc
@@ -1,0 +1,1 @@
+:current-version: 1.10
--- a/version.adoc
+++ b/version.adoc
@@ -12,9 +12,9 @@ or with the commandline switch `-a final`.
 // No changes needed beyond this line
 ifdef::final[]
 :current-version: {version}
-:attribute-version: {version}
+:current-version-as-attribute: {version}
 endif::[]
 ifndef::final[]
 :current-version: {version} draft
-:attribute-version: {version}-draft
+:current-version-as-attribute: {version}-draft
 endif::[]
--- a/version.adoc
+++ b/version.adoc
@@ -1,1 +1,20 @@
-:current-version: 1.10-draft
+// Set version
+:version: 1.10
+
+////
+Normally, a draft is produced, unless the `final` attribute is defined.
+Define the `final` attribute by uncommenting the following line
+or with the commandline switch `-a final`.
+////
+// :final:
+
+
+// No changes needed beyond this line
+ifdef::final[]
+:current-version: {version}
+:attribute-version: {version}
+endif::[]
+ifndef::final[]
+:current-version: {version} draft
+:attribute-version: {version}-draft
+endif::[]
--- a/version.adoc
+++ b/version.adoc
@@ -1,1 +1,1 @@
-:current-version: 1.10
+:current-version: 1.10-draft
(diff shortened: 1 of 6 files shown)


zklaus:
This single sources the version number as discussed in #343. There are a few open questions.

I split the version itself out into the new file `version.adoc`, which allows us to share the same version between `cf-conventions.adoc` and `conformance.adoc`.
There was one small hiccup with the conformance document, which is that it used the version even in the top level anchor.
The solution to address this here is to remove the version from the anchor, but this should be discussed.

One other idea is to add the "draft" bit as well, either in a kind of "long" or "full" version attribute, or as a separate `status-designator` attribute.

Finally, I made a few more replacements than originally mentioned. Most, I think, are uncontroversial, but please have a look and see if you disagree with any of them.

version.adoc
```diff
@@ -0,0 +1 @@
+:current-version: 1.10
```
erget, line 1: Might it make sense to do this?:

`:current-version: 1.10 draft`

And then remove ` draft` pre-release?

erget:
@zklaus this looks cleverly solved to me. @davidhassell any thoughts here? You can check [the artefacts](https://github.com/cf-convention/cf-conventions/pull/344/checks) and see how it looks when built.

I've made a suggestion on `version.adoc` that might improve handling, let me know what you think.

version.adoc
zklaus: I had this locally at some point, but I removed it again because I felt that it looked odd in the `:Conventions` attribute both in the examples, which  would read
```
// global attributes:
  :Conventions = "CF-1.10 draft" ;
  :featureType = "timeSeries" ;
```
and in the text where this attribute is described, for example
```
[...] attribute **`Conventions`** to a string value that contains "**`CF-1.10 draft`**".
```
Hence my suggestion to add a second `:status-indicator:` or similar that can be combined with the `:current-version:` where appropriate.

But I am happy to use the simpler solution :)

erget: I don't have overly strong feelings here, but in the absence of Sturm und Drang from other corners my preference would be to go with using `current-version` for the whole shebang. Reasoning:
- It's simpler than having 2 attributes
- It doesn't result in invalid CDL, since space characters are allowed
and thus it would be easier to maintain, and also more obvious that one is working with a draft.

zklaus:
I have added a fancified version of the version handling. Let me know what you think or if you want me to explain a bit more.

erget (approved):

erget (approved):
Very nice and smooth :)

zklaus:
Thanks, @erget. Before we merge, we should address the point raised by @ethanrd in #343, namely the name of the second attribute.

Also, there is one caveat: The automatic "final" tagging is hard to test because the real conditions only show up at the release, so at least for the first release we should be ready for it to fail and have an eye on that.

erget:
Would it be possible to make and destroy a test release? Or is that over the top?

zklaus:
I don't think that's necessary. Let's just check that everything works as expected. If it doesn't, we'll update the artifacts post-hoc.

PR #355, opened 2022-02-09
Update two mentions of 1.8 to 1.9

joshmoore:
Two locations mention that the value "CF-1.8" MUST be used for conforming datasets. Naively I assumed that these values should be updated "1.9". (Looking at previous commits like https://github.com/cf-convention/cf-conventions/commit/b6c65807a7dbf1eb78d6f711cd079296d7f43b63 this looks to be the case.)


zklaus:
Thanks for the report, @joshmoore. I think this is the same issue that has been discussed in #343 together with the solution in #344. Is this correct? Do those address the issue satisfactorily for you?

joshmoore:
Definitely. I looked for something that replace 1.9 but nothing as advanced as a variable replace. :+1: All the best.

PR #365, opened 2022-04-29
Fixes #362

### LINKED ISSUE #362: Provide additional guidance on moderating issues
Spawned from this item in #275:

> Jonathan: To record David’s suggestion of making a comment in the issue when the summary is updated, in order that everyone will be notified, and also because the moderator’s summary is part of the discussion and needs to be recorded therein for the discussion to make sense when read subsequently

Concretely I interpret this to mean that CONTRIBUTING.md should be updated to:
1. Specify that the moderator should summarise the discussion in an issue by updating the issue's description occasionally, and make a comment in the issue doing so. This ensures that everyone is notified, and records the moderator’s summary as part of the discussion for future reference
2. At the end of discussion, announce the merge date - and then merge on that date.

Additionally, if we adopt Kanban, this could signify a transition to a "ready for merge" column or something similar.

@JonathanGregory I take it this is in line with what you were thinking as well?

erget:
Fixes #362

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,6 +41,15 @@ All new pull requests should be submitted to the master branch of this repositor
 It is recommended that pull requests are created on a branch of a personal fork of this repository.
 Use of other branches is at the discretion of the repository administrators.
 
+## Moderating proposals
+The rules foresee a moderator for proposals.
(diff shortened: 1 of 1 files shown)


JonathanGregory (approved):
Thanks. That looks fine to me.

PR #376, opened 2022-08-22
master-->main

erget:
As agreed in https://github.com/cf-convention/discuss/issues/125


davidhassell (approved):
Looks good - thanks, Daniel.

PR #398, opened 2022-09-14
Apply line break policy

erget:
@sadielbartholomew happy with these?

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,5 +32,5 @@ The gh-pages branch is deployed to github pages at the following documents:
 - http://cfconventions.org/cf-conventions/conformance.html
 - http://cfconventions.org/cf-conventions/conformance.pdf 
 
-These documents are linked in [the cf-conventions web site](https://github.com/cf-convention/cf-convention.github.io) which is published to: http://cfconventions.org/.
+These documents are linked in [the cf-conventions web site](https://github.com/cf-convention/cf-convention.github.io) which is published to: [http://cfconventions.org/](http://cfconventions.org/).
 The latest spec documents are linked from those web pages here: http://cfconventions.org/latest.html. 
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ The latest build of the specification is held in the `gh-pages` branch and can b
 For the official web site please visit: http://cfconventions.org and the [corresponding GitHub organisation](https://github.com/cf-convention).
 
 # Contributing to the CF Conventions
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -22,12 +23,14 @@ To convert the AsciiDoc files into the resulting HTML file:
...
+Whenever a [pull request](https://github.com/cf-convention/cf-conventions/pulls) is merged, a [travis-ci build](https://travis-ci.org/github/cf-convention/cf-conventions) generates the latest specification draft and adds it to the [gh-pages branch here](https://github.com/cf-convention/cf-conventions/tree/gh-pages).
+The gh-pages branch is deployed to github pages at the following documents:
 - http://cfconventions.org/cf-conventions/cf-conventions.html
 - http://cfconventions.org/cf-conventions/cf-conventions.pdf
 - http://cfconventions.org/cf-conventions/conformance.html
 - http://cfconventions.org/cf-conventions/conformance.pdf 
 
-These documents are linked in the cf-conventions web site. [source here](https://github.com/cf-convention/cf-convention.github.io) which is published to: http://cfconventions.org/. The latest spec documents are linked from those web pages here: http://cfconventions.org/latest.html. 
\ No newline at end of file
+These documents are linked in [the cf-conventions web site](https://github.com/cf-convention/cf-convention.github.io) which is published to: http://cfconventions.org/.
```
sadielbartholomew, line 35: If we want precise URLs quoted (e.g. so printed documents state them), as I understand is the reason for not hyperlinking here (unless because it's linked closely elsewhere?), we can still hyperlink the same URL I would think, with the benefit of making it also clickable?

```suggestion
These documents are linked in [the cf-conventions web site](https://github.com/cf-convention/cf-convention.github.io) which is published to: [http://cfconventions.org/](http://cfconventions.org/).
```

README.md
```diff
@@ -22,12 +23,14 @@ To convert the AsciiDoc files into the resulting HTML file:
...
+The latest spec documents are linked from those web pages here: http://cfconventions.org/latest.html.
```
sadielbartholomew, line 36:
```suggestion
The latest spec documents are linked from those web pages here: [http://cfconventions.org/latest.html](http://cfconventions.org/latest.html). 
```

sadielbartholomew (approved):
All good, with a few related very minor suggestions made. But feel free to merge regardless of whether they are followed.

Issue #393, opened 2022-09-13
Cut all trailing whitespace including funky unicode characters


sadielbartholomew:
The AsciiDoc files, newly cleaned up as part of #387, still have trailing whitespace, including whitespace in the form of some funky unicode characters, e.g. `U+00A0`, which should be removed for code hygiene.

Suggested as part of abiding by style guides in https://github.com/cf-convention/cf-conventions/issues/373#issuecomment-1245219610. While the specific style guides are being debated, all (sensible ones) will advocate for no trailing whitespace, so this is not controversial and can be done immediately, and should be done as one, like with the first stage clean-up #387, to aid reviewing and avoiding creating merge conflicts.

Assigning myself to try to do this now.

sadielbartholomew:
Forgot to add a quick screenshot of my text editor with a whitespace-detecting mode enabled, demonstrating some of this on `ch09.adoc` as an example:

![whitespace_cf_conventions](https://user-images.githubusercontent.com/30274190/189953324-0553cad5-47e3-40e9-a8e6-2fdbaffd3078.png)

PR #399, opened 2022-09-14
#393 Whitespace fixes

squaregoldfish:
See issue #393 for discussion of these changes.

- All trailing whitespace (including conventional spaces and non-standard unicode-style spaces/non-breaking spaces) have been removed.
- All instances of the tab character have been replaced with two spaces, or an appropriate number to maintain alignment.
- The file `cf-conventions.html` has been added to the `.gitignore` file to reduce the risk of it being accidentally committed.


erget:
@squaregoldfish this looks fine to me, except that I agree with your suggestion of adding yourself to `conventions_contributors` rather than authors (at least for the moment). I've removed you from the author list but can't checkout the repo without having to do some distracting git stuff - would you mind adding yourself to the contributors list?

erget (approved):

Issue #164, opened 2019-06-12
Formatting errors in table of contents for examples.


martinjuckes:
**Title:** Formatting errors in table of contents for examples.
**Moderator:** @martinjuckes
**Requirement Summary:** (1) Numbering of index items is not consistent with target examples (e.g index entry `7.8. Climatological seasons` links to `Example 7.9. Climatological seasons`) , (2) examples 7.1 and 7.15 have no index entry, (3) Section 7.5 Geometries (including example 7.15) is included in the source document but omitted from the web page and pdf document.
**Technical Proposal Summary:** (1) Edit `toc-extra.adoc` to correct numbering of index items, (2) Correct an apparent error in defining the anchor for example 7.15 in `ch07.adoc`.

**Benefits:** Corrected index and inclusion of missing sub-section.
**Status Quo:** Index is incomplete and inacuurate.  
**Detailed Proposal:** 

- The numbering of the entries for examples 7.2 to 7.14 need to be corrected in `toc-extra.adoc` (currently set to 7.1 to 7.13);
- New entries need to be added to `toc-extra.adoc` for examples 7.1, 7.15 and 7.16 and for table 7.1;
- In `ch07.adoc` correct `[["timeseries-with-geometry"]]` to `[[timeseries-with-geometry]]`.
- The web page is more than a year behind the latest version in github .. missing Geometries section : is this intended?

To avoid potential conflict (unlikely) with formatting changes proposed in https://github.com/cf-convention/cf-conventions/pull/161, these changes are probably best tackled after that request is accepted.

JonathanGregory:
Dear @martinjuckes 

Thanks for raising this issue, the best part of four years ago, that identifies a number of errors. I suspect no-one responded at the time because of the possible interaction, which you highlighted, with https://github.com/cf-convention/cf-conventions/pull/161. If you have a few minutes available, it would be helpful if you could say whether any of these errors still remain now. Perhaps they were corrected by another issue at some point.

Best wishes and thanks

Jonathan

PR #524, opened 2024-06-06
Fix trac links

cofinoa:
This is a link fixing 

# Release checklist
- [] `history.adoc` up to date?

# For maintainers
Tags are set at the conclusion of the annual meeting; until then, `main` always is a draft for the next version.

### DIFF
--- a/history.adoc
+++ b/history.adoc
@@ -1,6 +1,6 @@
 :issues: https://github.com/cf-convention/cf-conventions/issues/
 :pull-requests: https://github.com/cf-convention/cf-conventions/pull/
-:tickets: https://cfconventions.org/Data/trac.html
+:tickets: https://cfconventions.org/Data/Trac-tickets/
 
 [[revhistory, Revision History]]
 == Revision History
--- a/appf.adoc
+++ b/appf.adoc
@@ -57,7 +57,7 @@ __Map coordinates:__:: The x (abscissa) and y (ordinate) rectangular coordinates
 __Notes:__:: Notes on using the **`PROJ`** software package for computing the mapping may be found at
 link:$$http://geotiff.maptools.org/proj_list/azimuthal_equidistant.html$$[http://geotiff.maptools.org/proj_list/azimuthal_equidistant.html]
 and
-link:$$http://proj.org/operations/projections/aeqd.html$$[https://proj.org/operations/projections/aeqd.html].
+link:$$https://proj.org/operations/projections/aeqd.html$$[https://proj.org/operations/projections/aeqd.html].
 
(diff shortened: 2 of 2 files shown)


cofinoa:
@sadielbartholomew I have fixed the broken links to the old Trac tickets.

Also, I have replaced an HTTP link with the correspondent working HTTPS

history.adoc
```diff
@@ -1,6 +1,6 @@
 :issues: https://github.com/cf-convention/cf-conventions/issues/
 :pull-requests: https://github.com/cf-convention/cf-conventions/pull/
-:tickets: https://cfconventions.org/Data/trac.html
+:tickets: https://cfconventions.org/Data/Trac-tickets/
```
sadielbartholomew, line 3: I might be missing some context, but the current link (with `trac.html`) works, and this suggestion doesn't, leading to a '404 - File not found' page when I check in the browser. Is there a corresponding page for Trac tickets that is due to be pushed from another branch to create this link, perhaps?

sadielbartholomew (commented):
Thanks @cofinoa, though I not sure why you are suggesting to move to this link which doesn't work at the moment - is there some context I have missed (sorry if it was raised in the Info Mgmt. meeting and I didn't pick up on it)?

history.adoc
cofinoa: Yes!, I have been very cryptic with this, it wasn't obvious to me neither. 

The link to the index of Trac tickets works:
> :tickets: https://cfconventions.org/Data/trac.html

but its purpose it's to define an attribute to be interpolated, as part of a complete link to the specific Trac ticket, later in the source file coded as, for example: 
> * {tickets}164.html[Ticket #164]: Add bounds attribute to first geometry CDL example.

Making all links to Trac tickets broken. Click them at::
https://cfconventions.org/Data/cf-conventions/cf-conventions-1.11/cf-conventions.html#_version_1_7_7_august_2017

cofinoa:
> Thanks @cofinoa, though I not sure why you are suggesting to move to this link which doesn't work at the moment - is there some context I have missed (sorry if it was raised in the Info Mgmt. meeting and I didn't pick up on it)?

history.adoc
sadielbartholomew: Aha, got it. Thanks for explaining, Antonio. That makes sense and I've checked some examples e.g. https://cfconventions.org/Data/Trac-tickets/164.html and they all work, so that's all good. Approving now.

sadielbartholomew (approved):
Working as tested. Thanks, please merge (or I can do that).

Issue #527, opened 2024-07-02
Boundary variables for auxiliary coordinates of more than two dimensions


JonathanGregory:
Section 7.1 on "Cell boundaries" contains the following text, about providing boundary variables for auxiliary coordinate variables of more than two dimensions:

> **Bounds for multi-dimensional coordinate variables with p-sided cells**

> In all other cases, the bounds should be dimensioned `(...,n,p)`, where `(...​,n)` are the dimensions of the auxiliary coordinate variables, and `p` the number of vertices of the cells. The vertices must be traversed anticlockwise in the lon-lat plane as viewed from above. The starting vertex is not specified.

There is no example given. I suppose it must mean something like this for three dimensions:

```
variables:
  float x(nk,nj,ni);
    x:bounds="x_bounds";
  float x_bounds(nk,nj,ni,p);
```

Since it talks about traversing the vertices clockwise in the lat-lon plane, it must be concerned with _horizontal_ faces of the 3D cells. A horizontal coordinate variable only has to be 3D (rather than 2D) if the cells aren't aligned in vertical columns. Do you agree with this interpretation? If so, we should clarify it. 3D cells don't just have horizontal faces, so this convention is of restricted use. For a more general treatment of bounds in two dimensions, we could refer to UGRID.

Cheers, Jonathan

JonathanGregory:
Dear all

I propose the following changes to remedy the defect described above in Sect 7.1 "Cell boundaries". In addition I propose to insert some new subsection headings, explanatory text, and rearrangement of text in Sect 7.1, to improve clarity. I don't believe this is any change to the meaning of the convention, except for one small generalisation (at the end). On account of that, and because the text changes are quite extensive, I've relabelled the issue as an enhancement.

The proposed changes are listed below. You might find it easier to look at the [pull request](https://github.com/cf-convention/cf-conventions/pull/547) or the [HTML](https://www.met.reading.ac.uk/~jonathan/CF_metadata/issue527/#cell-boundaries) of the modified document.

Cheers

Jonathan

* On re-reading, I suppose that "multidimensional" is referring to the dimensionality of the arrays, rather than the physical space. To clarify this, I've propose deleting this word, and adding a sentence to the end of the first paragraph of Sect 7.1: "CF can currently describe boundaries for cells which have one or two spatial dimensions, but does not provide conventions to describe the boundaries of cells with three spatial dimensions. Please refer to UGRID for development of such conventions."

[...]

PR #547, opened 2024-09-20
Clarify the conventions for boundary variables

JonathanGregory:
See issue #527 for discussion of these changes.

# Release checklist
- [Y] `history.adoc` up to date?


ch07.adoc
```diff
@@ -15,12 +15,18 @@ For cases when data pertain to geospatial features with highly variable geometry
...
-For grids constructed from cells that do not all have the same number of sides (e.g., a grid with some rectangular cells and some triangular cells), the vertex dimension must be at least as large as the maximum number of cell vertices.
+
+The vertex dimension must be of size two if the associated variable is one-dimensional (<<bounds-one-d>>), and of size greater than two if the associated variable has more than one dimension (<<bounds-lat-lon>>).
+For grids constructed from cells that do not all have the same number of sides (e.g., a grid with some rectangular cells and some triangular cells), the vertex dimension must be at least as large as the maximum number of cell vertices (<<bounds-two-d>>).
 For cells with fewer vertices than the size of vertex dimension, the unneeded elements must appear as the last elements in the vertex dimension and must be assigned the **`_FillValue`**.
+CF can currently describe boundaries for cells which have one or two spatial dimensions, but does not provide conventions to describe the boundaries of cells with three spatial dimensions.
+Please refer to <<UGRID>> for development of such conventions.
```
ChrisBarker-NOAA, line 29: UGRID only kinda-sorta does 3D -- not sure it should be mentioned here.

ChrisBarker-NOAA: But it probably should be mentioned somewhere near here ....

ChrisBarker-NOAA (approved):
looking good -- thanks!

TomLav:
Thanks for the PR @JonathanGregory. I will discuss a bit further below.

We remove the *reasonable assumption* and add the *no-default* statement. It strikes me that we have maybe not concluded on the corner-case when the data producer wants to convey that the cells are in fact points. Should we include the corner-case in the *no-default* disclaimer?

> Without a boundary variable, the extent of a cell is not known, nor whether adjacent cells are contiguous, separated by a gap, overlapping, or in fact points (cells with zero lengths).

ChrisBarker-NOAA:
>  It strikes me that we have maybe not concluded on the corner-case when the data producer wants to convey that the cells are in fact points.

What does "the cells are in fact points" mean? in my mind, points, are well, points, and not cells at all, and that's the usual definition of data in CF that doesn't have cell bounds.

the closest I've been able to find for a definition of "cell" is:

> "When gridded data does not represent the point values of a field but instead represents some characteristic of the field within cells of finite "volume,"...

"finite" to me means, "not a point".

taylor13:
I noticed the statement under figs. 7.1 and 7.2 that say:  "Tuples **`(lon(i),lat(j))`** represent grid cell centers."  Given the discussion, that the cell coordinate values aren't necessarily half way between the bounds, should these statements be reworded?  Perhaps "Tuples **`(lon(i),lat(j))`** represent the coordinate location of a cell."?  Or perhaps "Tuples **`(lon(i),lat(j))`** represent grid cell nominal centers."?

TomLav:
@ChrisBarker-NOAA @JonathanGregory 

I am starting to wonder if our idea of adding a "no-assumption" sentence in Chapter 7 is that good.

Re-reading the very start of section 7:
> When gridded data **does not represent the point values of a field** but instead represents some characteristic of the field within cells of finite "volume," a complete description of the variable **should include metadata that describes the domain or extent of each cell**, and the characteristic of the field that the cell values represent.

To me, this reads like the absence of `:bounds` should rightly be interpreted as providing locations for *point values of a field*. Because providing `:bounds` was specifically created for the cases where the data does not represent point values. Shouldn't the logical interpretation of the absence of `:bounds` rather be *point values*?

The new `no-assumption` sentence states that we cannot assume anything about cells in the absence of :bounds, but by doing so it actually brings the concept of a cell to mind, which seems contradictory to the first sentence of Chapter 7.
 
Maybe the best solution would be to delete the *reasonable assumption* sentence from Chapter 4, and add a sentence in Section 7 that reads: "In the absence of :bounds, the data represents the point values of a field". Then it is clear what the interpretation is when no :bounds are provided.

Sorry, I am going in circles a bit. Removing the sentence from Chapter 4 is still a very good idea.

taylor13:
I don't think the absence of bounds can imply "point" data unless we make bounds a *requirement* for data representing cells.  In the past, this has not been a requirement, so we can't change this without upsetting backward compatibility.

TomLav:
In that case, the absence of `:bounds` cannot imply anything: neither that the axis value represents cells, nor that such cells are of any extent and shape.

> Without a boundary variable, the axis values can neither be assumed to hold point positions, nor that the axis values represent cells. An unambiguous way to define point positions is to use the `:bounds` attribute to define 0-length cells. The only way to define cells, their position, and extent, is to use the `:bounds` attribute, as described below.

ChrisBarker-NOAA:
> I don't think the absence of bounds can imply "point" data unless we make bounds a requirement for data representing cells. In the past, this has not been a requirement, so we can't change this without upsetting backward compatibility.

Is that really the case? how in the world can you have cells if you haven't defined them somehow?

If it really was the case that that one could put in data representing cells without defining what the cells are, then I suppose what you had was: 

These data are on cells of unknown geometry -- seems like a bad idea to me, but if that's what CF used to allow, then I guess it still does. 

So how do you know if the data are point data or cell data? 

Is it point data if there is no cell_method, and cell data if there is?

in the current text:

> When gridded data does not represent the point values of a field but instead represents some characteristic of the field within cells of finite "volume," a complete description of the variable should include metadata that describes the domain or extent of each cell..and the characteristic of the field that the cell values represent

OK, so that is a "should" not a must....

But if it IS cell data, then it must (?) have a cell_method and/or a cell_measure (e.g. cell-area).

I guess it's not useless to have, e.g. have a cell_method and a cell-area with no defined bounds, though not great.

Even a cell_method with no other definition of the cells is still some information.

Not sure what this means for the text, but maybe something along the lines of, in the intro to 7, something like:

"Data is Representative of Cells if there is a cell_method or a cell metric defined"

Along with the "should" for the bounds.

JonathanGregory:
If you have comments (other than trivial ones on typos and words), please could you make them in #527, not here in the PR. Thanks.

ch07.adoc
```diff
@@ -2,25 +2,40 @@
...
+This is true for many types of climatological statistic; for example, the mean January temperature for the years 1971-2000 is computed from the 30 individual months of January, which are a set of discontiguous time-intervals.
+Climatological statistics are of such importance that we provide special methods for describing their associated computational domains in <<climatological-statistics>>.
+As an alternative to _n_-dimensional volumes with bounds, we provide <<geometries>>, for the case of geospatial applications in which each data value pertains to a single real-world feature, such as a river, watershed or country, represented by one or more points, lines or polygons.
+
 
 [[cell-boundaries, Section 7.1, "Cell Boundaries"]]
 === Cell Boundaries
 
-To represent cells we add the attribute **`bounds`** to the appropriate coordinate variable(s).
+To represent cells, the **`bounds`** attribute may be added to the appropriate coordinate variable(s).
```
ChrisBarker-NOAA, line 18: maybe "to define cells ..." rather than "represent"?  -- no biggie.

ChrisBarker-NOAA (commented):
Asside from one micro copy-edit suggestion, this looks good to go! 

Thanks for keeping up with the slog!

ch01.adoc
```diff
@@ -172,9 +174,17 @@ On the other hand identifying the vertical coordinate is not necessarily suffici
...
 
-It is often the case that data values are not representative of single points in time and/or space, but rather of intervals or multidimensional cells.
-This convention defines a **`bounds`** attribute to specify the extent of intervals or cells.
-When data that is representative of cells can be described by simple statistical methods, those methods can be indicated using the **`cell_methods`** attribute.
+It is often the case that data values are not representative of single points in time, space and other dimensions, but rather of intervals or multidimensional cells.
+CF defines a **`bounds`** attribute to specify the extent of intervals or cells.
+Because both the <<NUG>> and <<COARDS>> define coordinate variables but not cells or bounds, many applications assume that gridpoints are always located at the centers of their cells.
+This assumption does not hold in CF. If bounds are not provided, the location of the gridpoint within the cell is undefined, and nothing can assumed about the location and extent of the cell.
```
TomLav, line 180: missing word: and nothing can **be** assumed...

TomLav:
I support all these modifications and clarifications.

ch07.adoc
JonathanGregory: Thanks. I've replaced "represent" with "delimit".

ch01.adoc
JonathanGregory: Thanks. Done.

PR #598, opened 2025-03-04
move important rule (about the use of [over _type2_] outside the exam…

### LINKED ISSUE #597: In Chapter 7.3.3, important aspect of the convention are in the caption of example 7.7
# Title
In Chapter 7.3.3, important aspect of the convention are in the caption of example 7.7

# Moderator
Not assigned

# Requirement Summary
An important aspect of the convention for cell_methods (namely the use of `over`) are "hidden" in the caption to example 7.7. I suggest the paragraph is made part of the normal text, and appear before Example 7.7 (since Example 7.7 uses `over`).

# Technical Proposal Summary

Move this text from caption to just before Example 7.7.
> If the method is mean, various ways of calculating the mean can be distinguished in the cell_methods attribute with a string of the form "mean where type1 [over type2]". Here, type1 can be any of the possibilities allowed for typevar or type (as specified in the two paragraphs preceding above Example). The same options apply to type2, except it is not allowed to be the name of an auxiliary coordinate variable with a dimension greater than one (ignoring the possible dimension accommodating the maximum string length). A cell_methods attribute with a string of the form "mean where type1 over type2" indicates the mean is calculated by summing over the type1 portion of the cell and dividing by the area of the type2 portion. In particular, a cell_methods string of the form "mean where all_area_types over type2" indicates the mean is calculated by summing over all types of area within the cell and dividing by the area of the type2 portion. (Note that all_area_types is one of the valid strings permitted for a variable with the standard_name area_type.) If "over type2" is omitted, the mean is calculated by summing over the type1 portion of the cell and dividing by the area of this portion.

# Benefits
Readers of the convention will first read about the rule with `over` before being shown an example of how it can be used.

# Status Quo
Right now, a reader of the convention can overlook the rule with `over` (as I just did).

# Associated pull request
PR #598 

# Detailed Proposal
See above.

TomLav:
…ple box to make it more visible

See issue #597  for discussion of these changes.

# Release checklist
- [x] Conformance document up to date?

### DIFF
--- a/ch07.adoc
+++ b/ch07.adoc
@@ -431,6 +431,7 @@ These name(s) must be a subset of the standardised **`area_type`** strings.
 This convention can accommodate cases in which a method is applied to more than one area type and the result is stored in a single data variable (with a dimension which ranges across the various area types).
 It provides a convenient way to store output from land surface models, for example, since they deal with many area types within each surface gridbox (e.g., **`vegetation`**, **`bare_ground`**, **`snow`**, etc.).
 
+
 [[mean-surface-temperature-sensible-heat-flux]]
 [caption="Example 7.7. "]
 .Mean surface temperature over land and sensible heat flux averaged separately over land and sea.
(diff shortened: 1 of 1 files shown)


taylor13:
I think there was agreement in issue https://github.com/cf-convention/cf-conventions/issues/597 that this should be implemented, so I'm reopening.

TomLav:
Sorry, I see this is an old PR of mine that I had forgotten. Is there anything I should do to facilitate the merging?

JonathanGregory:
I think the issue was closed by accident. Thanks for noticing, Karl. You don't need to do anything, thanks, @TomLav. It looks ready to merge.

Issue #582, opened 2024-12-28
Conventions for anomalies


JonathanGregory:
## Moderator
None yet
## Requirement Summary
To provide conventions which describe the calculation of an anomaly (i.e. deviation) from the normal (i.e. reference or baseline) of the same geophysical quantity. The most important and complicated case is the calculation of anomalies with respect to climatological statistics.
## Status Quo
Sections 7.3 introduces `cell_methods`, and Section 7.4 defines the use of `cell_methods` to describe climatological statistics. There are various standard names of the form _X_`_anomaly`, where _X_ is another standard name. [Vocabulary issue 27](https://github.com/cf-convention/vocabularies/issues/27) proposes to add text in the descriptions of those standard names to say that a size-one coordinate variable with the standard name of `reference_epoch` can be used to record the time-bounds of a climatology with respect to which the anomaly was calculated. CF doesn't have any other conventions for describing anomalies or relating them to normals.
## Associated pull request
None yet.
## Background
This issue arises from three others:

* [vocabulary issue 27](https://github.com/cf-convention/vocabularies/issues/27) on "Adding reference epoch sentence to anomaly terms".

* [vocabulary issue 70](https://github.com/cf-convention/vocabularies/issues/70) (now closed, opened as `discuss` issue 252) on "reference periods for variables derived from climatology".

* [Discussion 305](https://github.com/orgs/cf-convention/discussions/305), entitled "Can CF offer better support for the representation of climate anomalies?"

There are some other unconcluded discussions relating to `cell_methods` too, which are more or less related to this one:

* [Discussion 372](https://github.com/orgs/cf-convention/discussions/372) (opened as issue #197 in this repo), entitled "Cell methods: `within`|`over` `days`|`months` and time axis (Section 7.4)" about the relationship of these to climatologies, and the possibility of `cell_methods` recording more than two statistical processing steps applied to the time axis.

* Issue #414 in this repo, proposing various clarifications to `cell_methods` and its documentation:

  * Reorder subsections 7.3.2 and 7.3.3.
  * Delete the default interpretation of `cell_methods` as `point` or `sum`.
  * Explicitly indicate that `where` can be used for the time dimension as well as spatial dimensions and that `where` can sometimes be interpreted as "when".
  * Revise some of the text and modify some of the examples for clarity and for readability.

* Issue #447 in this repo, on recording weighting in `cell_methods`. (This was raised in #414 and split off because of its greater urgency, and to make the discussion more manageable.)

[...]

PR #600, opened 2025-08-04
conventions for anomaly data

JonathanGregory:
See issue #582 for discussion of these changes.

# Release checklist
- [ ] `history.adoc` up to date?
- [ ] Conformance document up to date?


sethmcg:
Under this convention, a data variable is described as an anomaly by including in its `cell_methods` attribute a string of the form "_name_`: [`_name_`: `…​`] anomaly wrt` _norm_". This `cell_methods` entry tells us:

- that the data variable contains anomalies (_A_ in the notation of the start of this section) calculated from data (_P_) with the same dimensions, as the difference from the normal statistic named _norm_ (_N_).
- at which stage of data processing the anomaly _A = P - N_ was calculated. This is known because the entries in `cell_methods` appear in order of application.
- that _N_ was calculated from the variation of _P_ over dimensions identified by the _name_(s).

Each _name_ must be the name of an axis (i.e., a dimension and its corresponding coordinate variable, or a scalar coordinate variable) of the anomaly data variable _A_. We call these the "anomaly axes". They must have standard_name attributes. Usually there is only one anomaly axis, and usually it is a spatiotemporal axis.

For instance, for a data variable containing anomalies with respect to the zonal mean, _name_ identifies longitude as the anomaly axis e.g. "`longitude: anomaly wrt` _norm_". For an anomaly with respect to the area minimum, we need two names e.g. "`lat: lon: anomaly wrt` _norm_". As described in [Section 7.3, "Cell Methods"](#cell-methods), the combination of horizontal coordinate variables can alternatively be represented by the word area, thus "`area: anomaly wrt` _norm_". For an anomaly with respect to a statistic computed over time or climatological time, _name_ identifies the time axis of the anomaly data variable.

For each anomaly axis (although usually there is only one), _N_ has a coordinate variable or scalar coordinate variable that indicates the range(s) of coordinates over which the statistic _N_ was calculated from the variation of _P_.  We call these (scalar) coordinate variables the "norm coordinate variables".  For instance, if _N_ contains zonal means, it has a norm coordinate variable for longitude that indicates the range of longitudes over which the zonal mean was calculated.

There are two alternatives for _norm_. In both cases, _norm_ is an ancillary variable of the anomaly data variable ([Section 3.4, "Ancillary Data"](#ancillary-data)).

[cont'd]

sethmcg:
A norm data variable for a climatological statistic (in the sense of [Section 7.4, "Climatological Statistics"](#climatological-statistics)) has a norm coordinate variable that must have

- a `cell_methods` entry containing `within` and `over` keywords, and
- a `climatology` attribute to identify its boundary variable.

Example 7.15 shows how the `cell_methods` of the norm data variable defines the norm. By contrast, metadata following the simpler convention of [Section 7.5.3, "Temporal anomalies using anomaly standard names"](#temporal-anomalies) does not completely define the norm, as illustrated in Examples 7.20 and 7.21.

In Example 7.15, the norm coordinate variable of time has just one element. If the climatological time axis is multivalued, a norm metadata variable is required ([Section 7.5.2, "Anomalies with respect to a norm metadata variable"](#anomalies-norm-metadata)).

[cont'd]

sethmcg:
Example 7.15. Distinguishing temporal anomalies with different kinds of norm

The anomaly data variable `delta_tas(time,latitude,longitude)` contains daily maxima (along the `time` dimension) for the 16th-19th of July, 2023 of the anomaly in `air_temperature` with respect to the climatological mean of 1990-2019. `delta_tas` is the difference between the daily maxima contained in data variable `tas` and the time-mean `climatological_tas` of the entire 30-year period 1990-2019. Note that `tas` would not normally be included in the dataset; it is shown here for comparison with `delta_tas` and `climatological_tas`:

`dimensions:
  time=4;
variables:
  float delta_tas(time,latitude,longitude);
    delta_tas:standard_name="air_temperature";
    delta_tas:units="degC";
    delta_tas:units_metadata="temperature: difference";
    delta_tas:cell_methods="time: maximum time: anomaly wrt climatological_tas";
    delta_tas:coordinates="climatological_time";
    delta_tas:ancillary_variables="climatological_tas";
  float tas(time,latitude,longitude);
    tas:standard_name="air_temperature";
    tas:units_metadata="temperature: on_scale";
    tas:units="degC";
    tas:cell_methods="time: maximum";
  float climatological_tas(latitude,longitude);
    climatological_tas:standard_name="air_temperature";
    climatological_tas:units="degC";
    climatological_tas:units_metadata="temperature: on_scale";
    climatological_tas:coordinates="climatological_time";
    climatological_tas:cell_methods="climatological_time: mean";
  double time(time);
    time:standard_name="time";
    time:units="days since 2023-07-16";
    time:bounds="time_bounds";
    time:calendar="standard";
  double time_bounds(time,two);
  double climatological_time;
    climatological_time:standard_name="time";
    climatological_time:units="days since 1990-01-01";
    climatological_time:bounds="climatological_time_bounds";
    climatological_time:calendar="standard";
  double climatological_time_bounds(two);
data:
  time_bounds=0,1, 1,2, 2,3, 3,4;
  climatological_time_bounds=0,10957; // 1990-01-01, 2020-01-01
`

[cont'd]

sethmcg:
The `cell_methods` of `delta_tas` records that daily maxima were calculated first, then anomalies taken with respect to the long-term mean. (Although the result would have been the same if the order of operations was reversed.)

Another possibility is that the daily anomalies are calculated with respect to the 30-year July climatological mean, contained in `climatological_tas`. The metadata of `delta_tas` is the same in both cases. The two possibilities are distinguished by the `cell_methods` and time coordinates of `climatological_tas`:

```
dimensions:
  time=4;
  climatological_time=1;
variables:
  float delta_tas(time,climatological_time,latitude,longitude);
    delta_tas:standard_name="air_temperature";
    delta_tas:units="degC";
    delta_tas:units_metadata="temperature: difference";
    delta_tas:cell_methods="time: maximum time: anomaly wrt climatological_tas";
    delta_tas:ancillary_variables="climatological_tas";
  float tas(time,latitude,longitude);
    tas:standard_name="air_temperature";
    tas:units="degC";
    tas:units_metadata="temperature: on_scale";
    tas:cell_methods="time: maximum";
  float climatological_tas(climatological_time,latitude,longitude);
    climatological_tas:standard_name="air_temperature";
    climatological_tas:units="degC";
    climatological_tas:units_metadata="temperature: on_scale";
    climatological_tas:cell_methods="climatological_time: mean within years
      climatological_time: mean over years";
  double time(time);
    time:standard_name="time";
    time:units="days since 2023-07-16";
    time:bounds="time_bounds";
    time:calendar="standard";
  double time_bounds(time,two);
  double climatological_time(climatological_time);
    climatological_time:standard_name="time";
    climatological_time:units="days since 1990-01-01";
    climatological_time:climatology="climatological_time_bounds";
    climatological_time:calendar="standard";
  double climatological_time_bounds(climatological_time,two);
data:
  time_bounds=0,1, 1,2, 2,3, 3,4;
  climatological_time_bounds=181,10773; // 1990-07-01, 2019-08-01, i.e. July climatology
```

Equivalently, the `climatological_time` dimension could be omitted from `delta_tas` and `climatological_tas`, with the `climatological_time` variable instead being identified as a scalar coordinate variable in the coordinates attributes of these data variables.

[cont'd]

ch07.adoc
```diff
@@ -723,14 +725,538 @@ data:  // time coordinates translated to datetime format
...
+  double climatological_time;
+    climatological_time:standard_name="time";
+    climatological_time:units="days since 1990-1-1";
+    climatological_time:bounds="climatological_time_bounds";
+    climatological_time:calendar="standard";
+  double climatological_time_bounds(two);
+----
+If the daily anomalies are calculated with respect to the 30-year July climatological mean:
+----
+  float delta_tas(time,latitude,longitude);
```
davidhassell:
```suggestion
  float delta_tas(time,climatological_time,latitude,longitude);
```

ch07.adoc
```diff
@@ -723,14 +725,538 @@ data:  // time coordinates translated to datetime format
...
+
+
+[[anomalies-norm-data, Section 7.5.1, "Anomalies with respect to a norm data variable"]]
+==== Anomalies with respect to a norm data variable
+
+In this case, the word __norm__ in **`cell_methods`** is the name of the norm data variable, which must exist in the dataset.
+If either the anomaly data variable or the norm data variable has a **`standard_name`** attribute, it must __not__ be a standard name ending in **`_anomaly`**, and if they both have **`standard_name`** attributes, they must contain the same standard name.
+The anomaly data variable must name the norm data variable in its **`ancillary_variables`** attribute (<<ancillary-data>>), as well as in **`cell_methods`**, in order to indicate the link between them.
+
+The __norm__ data variable must have all the same coordinate variables (and hence dimensions) and scalar coordinate variables as the anomaly data variable, _except_ for the anomaly coordinate variables.
```
davidhassell: Is this right? Surely N shares only a subset of the dimensions and coordinate variables of A (as stated in the  second paragraph of "Anomaly Data")?

ch07.adoc
```diff
@@ -723,14 +725,538 @@ data:  // time coordinates translated to datetime format
...
+This is known because the entries in **`cell_methods`** appear in order of application.
+
+* that __N__ was calculated from the variation of __P__ over dimensions identified by the __name__(s).
+
+Each __name__ must be the name of a coordinate variable (and dimension) or a scalar coordinate variable of the anomaly data variable __A__.
+We call these the "anomaly coordinate variables".
+They must have **`standard_name`** attributes.
+Usually there is only one anomaly coordinate variable, and usually it is a spatiotemporal coordinate variable.
+
+For instance, for a data variable containing anomalies with respect to the zonal mean, __name__ identifies the coordinate variable or scalar coordinate variable of longitude as the anomaly coordinate variable e.g. "**`longitude: anomaly wrt`** __norm__".
```
davidhassell: Is there a particular reason why this isn't `anomaly_wrt` (i.e. replacing the space with an underscore). It's a single directive, I think, unlike `within` say, which can be followed by different terms (`days`, `years`). The underscore reinforces that, I feel (and makes it easier to parse, but that's not a bug deal).

sethmcg:
7.5.2.  Anomalies with respect to a norm metadata variable

In this case, the word _norm_ in `cell_methods` is the the name of the _norm metadata variable_. This is a variable in the dataset that is _not_ the norm data variable itself (although that may also be present in the dataset), but a variable that records the definition of the norm via the attributes `cell_methods` (which is mandatory) and `coordinates` (optional).  Like a grid mapping variable, it is of arbitrary type because it contains no data.  Although it is formally a data variable, CF recommends that the norm metadata variable not have any of the attributes of [Appendix A, Attributes](#attribute-appendix) other than `cell_methods` and `coordinates`. Attributes describing the norm quantity (standard name, units, etc.) are unnecessary, because they must be the same as for the anomaly data variable. The anomaly data variable must name the norm metadata variable in its `ancillary_variables` attribute ([Section 3.4, "Ancillary Data"](#ancillary-data)), as well as in `cell_methods`, in order to indicate the link between them.

For each anomaly axis, the norm metadata variable must have a coordinate variable with the same `standard_name` as the corresponding coordinate variable on the anomaly data variable.  These norm coordinate variables must be scalar or size-one, except in the case where _N_ has a climatological time axis as described in [[Section 7.4]](#climatological-statistics) (e.g., a monthly climatology).  Scalar norm coordinate variables must be named in the `coordinates` attribute of both the norm metadata variable and the anomaly data variable.  Size-one dimensions of the norm metadata variable must also be dimensions of the anomaly data variable.  The norm metadata variable must have no coordinate variables other than the norm coordinate variables.

[cont'd]

sethmcg:
If the norm has a multivalued climatological time axis, further information must be provided to describe the correspondence between elements of the anomaly and elements of the norm.  For example, in the case of an anomaly relative to a monthly climatology, all of the January anomaly values will be relative to the average value for January, all the February anomalies will be relative to the average value for February, and so on.

The mapping between the time axis of the anomaly variable and the climatological time axis of the norm is recorded by an auxiliary coordinate variable named by the `coordinates` attribute of the norm metadata variable.  This auxiliary coordinate variable is one-dimensional and has the same time dimension as the anomaly data variable.  The value of each element of the auxiliary coordinate variable is the index (numbering from 0) of the climatological time dimension of the norm corresponding to that element of the anomaly.  The norm metadata variable also has the same time dimension as the anomaly data variable (rather than being a scalar), but as usual, its values are arbitrary and meaningless.

The climatological time axis of `N` must be included in the file, although `N` itself need not be present.  The auxiliary coordinate variable is linked to this axis by a `compress` attribute naming the climatological time axis.  This method of mapping between axes is equivalent to [compression by gathering](#compession-by-gathering), described in Section 8.2, and using the same attribute for both means that the climatological time axis can be subset to only those values that are used.

In the following example, by "timestep `i`" we mean the slice of the data variable along its time dimension with index `i` (recalling that index numbering starts with 0).  Suppose that `A` contains monthly anomalies for water years 2011 and 2012 with respect to`N`, the 30-year climatological monthly means for 1980—2009.  `N` has a climatological time axis with a dimension of 12, one for each month January through December.  The anomaly axis is time, and `A` has a time dimension of 24 (two years times twelve months).  Water years run from September to September, so the first time coordinate of `A` is October 2010, and the first value of the auxiliary coordinate variable is 9, indicating that timestep 0 of `A` is the anomaly with respect to the timestep 9 of `N` (which is October for both).  Timestep 1 has an value of 10, and so on until timestep 3 (January), where the indexing for climatological time wraps around and the auxiliary coordinate variable has a value of 0.

In an abstract sense, the norm metadata variable indicates that `A` is the difference between `P` and `N'`, where `N'` has all same dimensions as `A` and `P`, including time, and is constructed by selecting, with repetition if necessary, and concatenating the slices of `N` whose climatological time indices are listed in the auxiliary coordinate variable.

Example 7.19 illustrates this convention, using the example described above.

[cont'd]

---

Comment: I had a hard time understanding this section, I think because the way it was originally written presumed proficiency with compression by gathering, which I've never used.  I've rewritten it pretty extensively to try to factor that out into a separate piece of the explanation, so please check that I didn't lose anything.

sethmcg:
Example 7.19. An anomaly data variable whose norm has a multivalued climatological time coordinate variable

The anomaly data variable delta_tas contains anomalies for the months of June, July and August of 2023 and 2024 with respect to the 30-year climatological means of those months for 1990—​2019. As well as the mandatory norm metadata variable climatological_tas_metadata, we show the norm data variable climatological_tas_data and the variable tas from which the anomalies werre calculated, although these are not normally present in the dataset, to illustrate the relationship between them.

```
dimensions:
  time=6;
  climatological_time=12;
variables:
  float delta_tas(time,latitude,longitude);
    delta_tas:standard_name="air_temperature";
    delta_tas:units="degC";
    delta_tas:units_metadata="temperature: difference";
    delta_tas:cell_methods="time: maximum time: anomaly wrt climatological_tas";
    delta_tas:ancillary_variables="climatological_tas_metadata";
  float tas(time,latitude,longitude);
    tas:standard_name="air_temperature";
    tas:units="degC";
    tas:units_metadata="temperature: on_scale";
    tas:cell_methods="time: maximum";
  int climatological_tas_metadata(time);
    climatological_tas_metadata:coordinates="month_indices";
    climatological_tas_metadata:cell_methods=
      "climatological_time: mean within years
      climatological_time: mean over years";
  int month_indices(time);
    month_indices:compress="climatological_time";
  double time(time);
    time:standard_name="time";
    time:units="days since 2023-06-01";
    time:bounds="time_bounds";
    time:calendar="standard";
  double time_bounds(time,two);
  double climatological_time(climatological_time);
    climatological_time:standard_name="time";
    climatological_time:units="days since 1990-01-01";
    climatological_time:bounds="climatological_time_bounds";
    climatological_time:calendar="standard";
  double climatological_time_bounds(climatological_time,two);
  float climatological_tas_data(climatological_time,latitude,longitude);
    climatological_tas_data:standard_name="air_temperature";
    climatological_tas_data:units="degC";
    climatological_tas_data:units_metadata="temperature: on_scale";
    climatological_tas_data:cell_methods=
      "climatological_time: mean within years
      climatological_time: mean over years";
data:
  time=15, 45, 76, 381, 411, 442;
    // 2023-06-16, 2023-07-16, 2023-08-16, 2024-06-16, 2024-07-16, 2024-08-16
  time_bounds=0,30, 30,61, 61,92, 366,396, 396,427, 427,458;
    // beginning and end of Jun, Jul and Aug of 2023 and 2024
  climatological_time=15, 45, ... 349;  // 1990-01-16, 1990-02-15 ... 1990-12-16
  climatological_time_bounds=0,10623, 31,10651, ... 334,10957;
    // 1990-01-01,2019-02-01, 1990-02-01,2019-03-01 ... 1990-12-01,2020-01-01
  month_indices=5, 6, 7, 5, 6, 7;
```

Element **0** of `month_indices` is **5**. This means that element **5** of `climatological_time`, which is the climatology for June 1991-2000, corresponds to element 0 of `time`, which is 2023-06-16, with bounds of 2023-06-01 and 2023-07-01 i.e. the whole of June 2023. It indicates that `delta_tas(0,:,:`), where by `:` we mean the entire range of the dimension, is the difference between `tas(0,:,:)` and `climatological_tas(5,:,:)`.

sethmcg:
7.5.3. Temporal anomalies using anomaly standard names

Several CF standard names ending with `_anomaly` have been defined for anomalies with respect to a long-term or climatological time-mean. The start and end of the climatological normal period may optionally be recorded in the bounds of a variable with a standard name of `reference_epoch`. This variable may be either a scalar coordinate variable, as in Example 7.20, or a coordinate variable with a single size-one dimension. This convention using `_anomaly` standard names and `reference_epoch` is restricted to those common cases that have such standard names defined.

Example 7.20. An anomaly data variable with a reference epoch
```
variables:
  float delta_tas(time,latitude,longitude);
    delta_tas:standard_name="air_temperature_anomaly";
    delta_tas:units="degC";
    delta_tas:units_metadata="temperature: difference";
    delta_tas:coordinates="reference_epoch";
    delta_tas:cell_methods="time: maximum";
  double time(time);
    time:standard_name="time";
    time:units="days since 2023-07-16";
    time:bounds="time_bounds";
    time:calendar="standard";
  double time_bounds(time,two);
  double reference_epoch;
    reference_epoch:standard_name="reference_epoch";
    reference_epoch:units="days since 1990-01-01";
    reference_epoch:bounds="reference_epoch_bounds";
    reference_epoch:calendar="standard";
  double reference_epoch_bounds(two);
data:
  time_bounds=0,1, 1,2, 2,3, 3,4;
  reference_epoch_bounds=0,10957; // 1990-01-01, 2020-01-01
```

The data variable `delta_tas` contains daily maximum temperatures for 16th-19th July 2023 expressed as anomalies with respect to a long-term time-mean. Note that `delta_tas` has the attribute `units_metadata="temperature: difference"` because it is the difference between two temperatures, as recommended in [Section 3.1.2, "Temperature units"](#temperature-units). The climatological reference period is defined as 1990-2019 by the bounds of `reference_epoch` (10957 days since 1st January 1990 in the standard calendar is 1st January 2020). The single value of the `reference_epoch` coordinate variable should be a representative time within the climatological interval (see [Section 7.4, "Climatological Statistics"](#climatological-statistics)).

In this example, `reference_epoch` is a scalar coordinate variable. Alternatively, it could be defined with a dimension `reference_epoch=1`, a one-dimensional coordinate variable `reference_epoch(reference_epoch)` and bounds `reference_epoch_bounds(reference_epoch,two)`. In this case, `reference_epoch` must be included among the dimensions of `delta_tas` e.g. `delta_tas(reference_epoch,time,latitude,longitude)`, but the `coordinates` attribute is not needed, unlike in the case of a scalar coordinate variable.

appe.adoc
```diff
@@ -19,6 +19,7 @@ This is the default method for a quantity that is intensive with respect to the
 | The data values are representative of a sum or accumulation over the cell.
 This is the default method for a quantity that is extensive with respect to the specified dimension.
 
+| `anomaly wrt` | __u__ | Difference from a statistical norm (<<anomalies>>).
```
davidhassell: Just wondering if it was considered to have an underscore in this: `anomaly_wrt`, as none of the other method names have a space in them (think of the poor parsers!).

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+
+In the case where __N__ has a multivalued climatological time axis (as described in <<climatological-statistics>> and illustrated in Examples 7.9, 7.10, and 7.11), the norm metadata variable has, as its sole dimension, the anomaly time dimension of __A__.
+In this case, the norm metadata variable can have more than one element.
+
+In all cases, norm coordinate variables must have boundary variables that indicate the coordinate ranges over which __N__ was calculated from __P__.
+The norm metadata variable must have a **`cell_methods`** attribute (<<cell-methods>>) to indicate how __N__ was computed from the variation of _P_.
+The **`cell_methods`** must contain an entry referring to the norm coordinate variable (or combination thereof, if more than one), or more than one entry if they contain **`within`** and **`over`** keywords, and must have no other entries.
+
+The use of norm metadata variables is illustrated by Examples 7.16, 7.17 and 7.18.
+The treatment of multivalued climatological time is described and illustrated after Example 7.18.
```
davidhassell:
```suggestion
The treatment of multivalued climatological time is described and illustrated after Example 7.19.
```

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+dimensions:
+  time=6;
+  climatological_time=12;
+variables:
+  float delta_tas(time,latitude,longitude);
+    delta_tas:standard_name="air_temperature";
+    delta_tas:units="degC";
+    delta_tas:units_metadata="temperature: difference";
+    delta_tas:cell_methods="time: maximum time: anomaly wrt climatological_tas";
+    delta_tas:ancillary_variables="climatological_tas_metadata";
```
davidhassell: I'm not sure what's going on, here. `climatological_tas` does not appear elsewhere in the example, and `climatological_tas_metadata` does not appear in the cell methods ...

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+data:
+  time_bounds=0,1, 1,2, 2,3, 3,4;
+  climatological_time_bounds=181,10773; // 1990-07-01, 2019-08-01, i.e. July climatology
+----
+Equivalently, the **`climatological_time`** dimension could be omitted from **`delta_tas`** and **`climatological_tas`**, with the **`climatological_time`** variable instead being identified as a scalar coordinate variable in the **`coordinates`** attributes of these data variables.
+====
+
+
+[[anomalies-norm-metadata, Section 7.5.2, "Anomalies with respect to a norm metadata variable"]]
+==== Anomalies with respect to a norm metadata variable
```
davidhassell, line 923: How can you tell that a size 1 norm data variable is not a norm metadata variable?

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+In Example 7.15, the norm coordinate variable of time has just one element.
+If the climatological time axis is multivalued, a norm metadata variable is required (<<anomalies-norm-metadata>>).
+
+[[reference-epoch-distinguishing-ex]]
+[caption="Example 7.15. "]
+.Distinguishing temporal anomalies with different kinds of norm
+====
+The anomaly data variable **`delta_tas(time,latitude,longitude)`** contains daily maxima (along the **`time`** dimension) for 16th-19th July 2023 of the anomaly in **`air_temperature`** with respect to the climatological mean of 1990-2019.
+**`delta_tas`** is the difference between the daily maxima contained in data variable **`tas`**, and the time-mean  **`climatological_tas`** of the entire 30-year period 1990-2019.
+Note that **`tas`** would not normally be included in the dataset; it is shown here for comparison with **`delta_tas`** and **`climatological_tas`**:
```
davidhassell: This implies to me that when `tas` is in the dataset, you should be able to infer its relationship to to `delta_tas` by inspecting the metadata, which you can't. How about:
```suggestion
**`delta_tas`** (A) is the difference between the daily maxima contained in data variable **`tas`** (P), and the time-mean  **`climatological_tas`** (N) of the entire 30-year period 1990-2019.
Note that **`tas`** has no metadata that formally identifies it as P, and it would not normally be included in the dataset; it is shown here for comparison with **`delta_tas`** and **`climatological_tas`**.
```

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+    delta_tas:units="degC";
+    delta_tas:units_metadata="temperature: difference";
+    delta_tas:cell_methods="time: maximum time: anomaly wrt climatological_tas";
+    delta_tas:coordinates="climatological_time";
+    delta_tas:ancillary_variables="climatological_tas";
+  float tas(time,latitude,longitude);
+    tas:standard_name="air_temperature";
+    tas:units_metadata="temperature: on_scale";
+    tas:units="degC";
+    tas:cell_methods="time: maximum";
```
davidhassell: Do we need to make the comparison again?
```suggestion
```

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+    time:calendar="standard";
+  double time_bounds(time,two);
+  double climatological_time;
+    climatological_time:standard_name="time";
+    climatological_time:units="days since 1990-01-01";
+    climatological_time:bounds="climatological_time_bounds";
+    climatological_time:calendar="standard";
+  double climatological_time_bounds(two);
+----
+If the daily anomalies are calculated with respect to the 30-year July climatological mean:
```
davidhassell, line 985: Either `delta_tas` has to have a `climatological_time` dimension, or the `climatological_time` variable must be a scalar ... the opposite of which are currently the case. Similar issues occur in other examples.

ch07.adoc
davidhassell: That said, I would just be happier not including it all - it confuses me more than helps :)

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+It would be constructed by selecting, with repetition if necessary, the slices of __N__ whose climatological time indices are listed in the auxiliary coordinate variable, then concatenating them.
+
+Example 7.19 illustrates this convention, using the example described above.
+
+[[anomaly-with-multivalued-climatology-ex]]
+[caption="Example 7.19. "]
+.An anomaly data variable whose norm has a multivalued climatological time coordinate variable
+====
+The anomaly data variable **`delta_tas`** contains anomalies for the months of June, July and August of 2023 and 2024 with respect to the 30-year climatological means of those months for 1990--2019.
+As well as the mandatory norm __metadata__ variable **`climatological_tas_metadata`**, we show the norm __data__ variable **`climatological_tas_data`** and the variable **`tas`** from which the anomalies werre calculated, although these are not normally present in the dataset, to illustrate the relationship between them.
```
davidhassell: As before, I would be happier to not include the non-connected variables.

davidhassell (commented):
A few comments ... thanks.

appe.adoc
JonathanGregory: Yes, good point. There are some other multi-word methods too (with `_`). I have changed it.

ch07.adoc
JonathanGregory: Sorry about the confusion. I have reworded the introduction to the example thus:

> The anomaly data variable (_A_, **`delta_tas`**) contains daily maxima (along its **`time`** dimension) for 16th-19th July 2023 of the anomaly in **`air_temperature`** with respect to the climatological mean (_N_, **`climatological_tas`**) of the 30-year period 1990-2019. In this example, we include the variable **`tas`** (_P_), from which _A_ was calculated, as _P_ &minus; _N_. The variable **`tas`** has no metadata that formally identifies it as _P_, and _P_ would not usually be included in the dataset; it is shown here for comparison of its metadata with **`delta_tas`** and **`climatological_tas`**:

and in the example I have put the variables in this order, with comments as shown:

```
  float delta_tas(time,latitude,longitude); // anomaly data variable A
  float climatological_tas(latitude,longitude); // norm data variable N
  float tas(time,latitude,longitude); // P for comparison, not usually included in dataset
```

ch07.adoc
JonathanGregory: Deleted

ch07.adoc
JonathanGregory: You can't, as things stand, in the unusual circumstance when the anomaly data variable has no coordinate variables other than anomaly coordinate variables (hence the norm data variable has no dimensions greater than 1). This had occurred to me, but I thought it didn't matter, but really it does, because you'd like to know whether the data in the variable is meaningful. Thanks for pointing it out. I will add an extra requirement (except in the case of climatological time) that a norm metadata variable, which is always of size 1, should contain missing data. A norm _data_ variable whose sole element was missing data would be no more use than a norm metadata variable anyway.

ch07.adoc
davidhassell: Very nice - thanks.

ch07.adoc
JonathanGregory: I have added:

> The norm metadata data variable is a "dummy" variable, like a grid mapping variable, for instance. Its type is arbitrary. It must have a **`_FillValue`** attribute, and its elements must be equal to the **`_FillValue`**, to indicate that it contains no meaningful data. (Under some circumstances, this condition alone distinguishes a norm metadata variable from a norm data variable.)

ch07.adoc
JonathanGregory: 7.18 is correct but I see how this could be misleading. I have changed it to, "The treatment of multivalued climatological time is described after Example 7.18 and illustrated in Example 7.19."

ch07.adoc
JonathanGregory: I don't understand. The `climatological_time` variable is a scalar, and named by `coordinates` of the anomaly data variable `delta_tas` and the norm data variable `climatological_tas`. What have I missed?

ch07.adoc
JonathanGregory: I've removed them.

ch07.adoc
JonathanGregory: Yes, it's a subset. It is all of them _except_ for the anomaly coordinate variables. How could this be written more clearly?

ch07.adoc
JonathanGregory: I've changed it to `anomaly_wrt`. Thanks.

ch07.adoc
JonathanGregory: That was a mistake, now fixed. Thanks.

JonathanGregory:
The changes I've made in response to David's comments can be seen at https://github.com/JonathanGregory/cf-conventions/compare/26fff26..60a3570

ch07.adoc
```diff
@@ -723,14 +725,536 @@ data:  // time coordinates translated to datetime format
...
+  float zmlongitude(zmlongitude);
+    zmlongitude:standard_name="longitude";
+    zmlongitude:units="degrees_E";
+  float zmbounds(zmlongitude,two);
+data:
+  zmlongitude=-165; // a value within the range; it's usually arbitrary
+  zmbounds=-240,-90;
+----
+
+The **`zmbounds`** indicate that **`rtoa`** contains anomalies with respect to the zonal mean calculated from 120&deg;E to 90&deg;W.
```
TomLav: Is this zonal mean calculated over the antimeridian, which is why 120E is recorded as -240? If the case, we could add this:

> with respect to the zonal mean calculated from 120&deg;E to 90&deg;W over the antimeridian.

If this is not the reason, why encode 120E as -240 ?

ch07.adoc
```diff
@@ -723,14 +725,536 @@ data:  // time coordinates translated to datetime format
...
+
+In the following example, by "timestep __i__" we mean the slice of the data variable along its time dimension with index __i__ (recalling that index numbering starts with 0).
+Suppose that __A__ contains monthly anomalies for the months of June, July, and August of 2023 and 2024 with respect to __N__, the 30-year climatological monthly means for 1990&ndash;2019.
+__N__ has a climatological time axis with a dimension of 12, one for each month January through December.
+The anomaly axis is time, and __A__ has a time dimension of 6 (two years times three months).
+The first time coordinate of __A__ is June 2023, and the first value of the auxiliary coordinate variable is 5, indicating that timestep 0 of __A__ (June 2023) is the anomaly with respect to the timestep 5 of __N__ (climatological June).
+It would also be valid to use compression by gathering and provide only the climatology for June, July, and August, in which case the climatological time axis would have a dimension of 3.
```
TomLav: This was not clear for me. What would change if compression by gathering was used? `climatological_time` would be length 3, but would the indexes stored in `month_indices` change to 0,1,2,0,1,2 as well? And would anything else change to indicate the compression by gathering was used?

Also "provide only the climatology for June, July, and August" is maybe unclear since when we use the norm metadata variable, we do not provide any climatology (data) in the first place.

ch07.adoc
TomLav: I am struggling to see if this `_FillValue` should also be added to some of the Examples.

ch07.adoc
```diff
@@ -723,14 +725,536 @@ data:  // time coordinates translated to datetime format
...
 
-[[geometries, Section 7.5, "Geometries"]]
+
+[[anomalies, Section 7.5, "Anomaly data"]]
+=== Anomaly data
+
+An "anomaly" is the difference between a physical quantity and its statistical norm.
+For example, a commonly-used anomaly is the current temperature at a specific location minus the long-term average temperature there.
+
+CF offers two conventions for describing anomaly data.
```
TomLav, line 735: Personally, I would prefer it if we could first mention the more general convention. Then the simple (can we call it "legacy") convention with standard name. This will give the general convention the focus it deserves.

ch07.adoc
```diff
@@ -723,14 +725,536 @@ data:  // time coordinates translated to datetime format
...
+====
+
+
+[[temporal-anomalies, Section 7.5.3, "Temporal anomalies using anomaly standard names"]]
+==== Temporal anomalies using anomaly standard names
+
+Several CF standard names (those ending with **`_anomaly`**) have been defined for anomalies with respect to a long-term or climatological time-mean.
+The start and end of the climatological normal period may optionally be recorded in the bounds of a variable with a standard name of **`reference_epoch`**.
+This variable may be either a scalar coordinate variable, as in Example 7.20, or a coordinate variable with a single size-one dimension.
+This convention using **`_anomaly`** standard names and **`reference_epoch`** is restricted to those common cases that have such standard names defined.
```
TomLav: Would it be natural to add: "and is considered a legacy solution" ?

ch07.adoc
sethmcg: I struggle with how to rewrite these sentences to put the general convention first, since most of what we have to say about is in in contrast to the simpler convention.  This is just the intro; I think the fact that we then proceed to talk a lot about the general convention first will suffice to put the focus on it.

I support the idea of calling the simpler convention "legacy".

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+
+The generalized definition of an anomaly value __A__ of some physical quantity __q__ is the difference __P__ &minus; __N__ between a particular value __P__ of __q__ and a normal value or norm __N__ of __q__.
+__N__ is some statistic calculated from the values of __q__ that lie within specified ranges of one or more of its coordinates.
+__P__ can be, but is not necessarily, one of the set of values from which __N__ is calculated.
+
+In the same way, a data variable __A__ containing anomalies (an "anomaly data variable") is notionally the difference between a data variable __P__ containing the original data and a data variable __N__ (a "norm data variable") containing the statistical norm.
+__P__ is typically not present in the dataset, and __N__ is usually absent as well.
+
+__P__ has all the same netCDF dimensions and coordinate variables as __A__, but __N__ shares only a subset of the dimensions and coordinate variables of __A__.
+Corresponding to each of the dimensions of __A__ which __N__ does not share, it has its own dimension for the same spatiotemporal coordinate or other quantity.
```
sethmcg: I cannot parse this sentence, nor the one that follows.

My suggestion was "The other dimensions of N are the ones over which the norm is calculated."  What is missing about that?  How does it need to change?

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+__N__ is most often either a time-mean over a continuous period of multiple years or a climatological time-mean (<<climatological-statistics>>).
+The time coordinate of the anomaly may or may not lie within the range of times from which __N__ is calculated.
+__N__ has all the same dimensions and coordinate variables as __A__ except for time.
+
+In general, a statistical norm can be calculated from any single dimension or combination of dimensions, and the norm statistic does not have to be a mean.
+For example, anomalies might be calculated (as a function of longitude) with respect to the zonal mean, or (as a function of horizontal location) with respect to the minimum value in the area.
+In these examples, the norm is the zonal mean or the area minimum, respectively.
+When temporal anomalies are described following the convention of this section, more information can be recorded about the norm than when following the alternative, less general, convention of <<temporal-anomalies>>.
```
sethmcg: Should we change "alternative, less general," to "legacy"?

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+In both cases, __norm__ is an ancillary variable of the anomaly data variable (<<ancillary-data>>).
+
+If the norm data variable __N__ is present in the dataset, it can be named as __norm__ in the **`cell_methods`** of the anomaly variable.
+Such would be the case in a dataset that contains both zonal means and anomalies relative to those means.
+No modification to the metadata of __N__ is required for it to serve as the norm for __A__, and __N__ may still be treated as a data variable in its own right as well.
+We describe the use of a norm data variable first (<<anomalies-norm-data>>) because this case is conceptually more obvious, although it is uncommon for __N__ to be present in the dataset.
+
+The second alternative (<<anomalies-norm-metadata>>) is where __norm__ in **`cell_methods`** identifies a "norm metadata variable" instead of the norm data variable.
+A norm metadata variable contains information about the norm axes, but no data of its own.
+This method can be used regardless of whether the norm data variable is present in the dataset as well.
```
sethmcg: "present in the dataset as well" sounds weird to me.  I would prefer "also present in the dataset."

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+The anomaly data variable must name the norm data variable in its **`ancillary_variables`** attribute (<<ancillary-data>>), as well as in **`cell_methods`**, in order to indicate the link between them.
+
+The __norm__ data variable __N__ must have all the same axes as the anomaly data variable __A__, __except__ for the anomaly axes.
+For each anomaly axis (although usually there is only one), the norm data variable has a coordinate variable (and dimension of the same name) or a scalar coordinate variable (named in its **`coordinates`** attribute).
+In either case, we refer to it as a "norm coordinate variable".
+The norm coordinate variable must have the same **`standard_name`** as the anomaly coordinate variable.
+It must also have boundary variables to indicate the coordinate range over which __N__ was calculated from __P__.
+
+Norm coordinate variables cannot have a dimension greater than 1, and these dimensions must be included among the dimensions of the anomaly data variable as well being dimensions of the norm data variable.
```
sethmcg: Typo: "as well being" -> "as well as being"

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+
+
+[[anomalies-norm-data, Section 7.5.1, "Anomalies with respect to a norm data variable"]]
+==== Anomalies with respect to a norm data variable
+
+In this case, the word __norm__ in **`cell_methods`** is the name of the norm data variable, which must exist in the dataset.
+If either the anomaly data variable or the norm data variable has a **`standard_name`** attribute, it must __not__ be a standard name ending in **`_anomaly`**, and if they both have **`standard_name`** attributes, they must contain the same standard name.
+The anomaly data variable must name the norm data variable in its **`ancillary_variables`** attribute (<<ancillary-data>>), as well as in **`cell_methods`**, in order to indicate the link between them.
+
+The __norm__ data variable __N__ must have all the same axes as the anomaly data variable __A__, __except__ for the anomaly axes.
```
sethmcg: I think 'norm' should not be italicized here.

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+For each anomaly axis (although usually there is only one), the norm data variable has a coordinate variable (and dimension of the same name) or a scalar coordinate variable (named in its **`coordinates`** attribute).
+In either case, we refer to it as a "norm coordinate variable".
+The norm coordinate variable must have the same **`standard_name`** as the anomaly coordinate variable.
+It must also have boundary variables to indicate the coordinate range over which __N__ was calculated from __P__.
+
+Norm coordinate variables cannot have a dimension greater than 1, and these dimensions must be included among the dimensions of the anomaly data variable as well being dimensions of the norm data variable.
+Likewise, any scalar norm coordinate variables must be named in the **`coordinates`** attribute of the anomaly data variable as well as the norm data variable.
+
+The __norm__ data variable must have a **`cell_methods`** attribute with an entry for the norm coordinate variable (or the combination of them if more than one) to indicate how __N__ was computed from the variation of __P__.
```
sethmcg: I think norm should not be italicized here, either.

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+For instance, the __norm__ data variable for anomalies with respect to a time-mean must have a norm coordinate variable for time, and a **`cell_methods`** attribute containing an entry naming this variable.
```
sethmcg: Nor should norm be italicized here, IMO.

ch07.adoc
```diff
@@ -723,14 +725,554 @@ data:  // time coordinates translated to datetime format
...
+Alternatively, it could be defined with a dimension **`reference_epoch=1`**, a one-dimensional coordinate variable **`reference_epoch(reference_epoch)`** and bounds **`reference_epoch_bounds(reference_epoch,two)`**.
+In this case, **`reference_epoch`** must be included among the dimensions of **`delta_tas`** e.g. **`delta_tas(reference_epoch,time,latitude,longitude)`**, but the **`coordinates`** attribute is not needed, unlike in the case of a scalar coordinate variable.
+====
+
+This convention, using **`_anomaly`** standard names and **`reference_epoch`**, is simpler than using a norm data or metadata variable, but provides no information about the nature of the norm, which is described by its **`cell_methods`** attribute.
+The norm data variable is not necessarily present in the file, and even if it is present, this approach does not provide any link between the anomaly variable and the norm variable.
+Hence, interpretation of the anomaly can be unclear.
+Example 7.21 illustrates this ambiguity as it manifests in Example 7.20.
+Such ambiguities can be resolved by the conventions of <<anomalies-norm-data>> and and <<anomalies-norm-metadata>>.
```
sethmcg: Typo: "and and"

ch07.adoc
TomLav: An attempt below (I do not insist on the two bullet points).

> CF offers two conventions for describing anomaly data.
> * In the remainder of this section, and the following two (<<anomalies-norm-data>> and <<anomalies-norm-metadata>>), we describe a general convention that can handle anomalies over any type of coordinate and allows describing in details how the anomaly and the associated norm were calculated.
> * In <<temporal-anomalies>>, we describe a legacy convention that relies on special standard names, handles only anomalies wrt the time dimension, and is less descriptive.

ch07.adoc
JonathanGregory: That is the reason, yes. I have changed it thus:

>  &nbsp;&nbsp;`zmlongitude=-165; // midpoint of the range`<br>  &nbsp;&nbsp;`zmbounds=-240,-90; // eastward, from 120E to 90W`

> The **`zmbounds`** indicate that **`rtoa`** contains anomalies with respect to the zonal mean calculated over longitudes ranging eastward from 120&deg;E to 90&deg;W. The **`zmlongitude`** should be a value within the range of longitudes covered, but its precise choice is usually arbitrary.

It's good that you made this point. People quite often ask about the choice of the coordinate in cases where it's the bounds they're really interested in. We usually reply that it's not prescribed, and you can choose whatever makes sense for the dataset concerned. This example shows that it's important for the coordinate to be _within the range of the bounds_. When the dimension has size 1, and the coordinate variable is cyclic, the coordinate value is the only way to be sure which way round the bounds should be interpreted. I wasn't aware of this. Maybe we've discussed it before!

ch07.adoc
davidhassell: Apologies - my mistake (I somehow confused climatological_tas with climatological_time) - this is fine!

ch07.adoc
```diff
@@ -723,14 +725,536 @@ data:  // time coordinates translated to datetime format
...
+  float climatological_tas; // norm metadata variable
+    climatological_tas:coordinates="climatological_time";
+    climatological_tas:cell_methods="climatological_time: mean";
+  double time(time); // anomaly coordinate variable
+    time:standard_name="time";
+    time:units="days since 2023-07-16";
+    time:bounds="time_bounds";
+    time:calendar="standard";
+  double time_bounds(time,two);
+  double climatological_time; // norm coordinate variable
```
davidhassell, line 978:
```suggestion
  double climatological_time; // norm and anomaly coordinate variable
```

ch07.adoc
davidhassell: Sorry - I un-apologise :) The problem is in the CDL below (this comment is on the caption for that)

ch07.adoc
JonathanGregory: I agree, it's not clear, because in the case with a climatological time dimension of 12 we're still using the mechanism of compression by gathering, although we're not actually compressing. I have replaced the sentence "It would also be valid ..." with

> Since only June, July, and August climatological means are needed in this example, we could alternatively give the climatological time axis of _N_ a dimension of 3, with elements for those three months alone. In that case, the first value of the auxiliary coordinate variable would be 0 for climatological June.

and added the following to the end of the example:

> If we chose instead to show the climatological time axis with only the three months needed in this example (June, July, and August), the following lines would replace the corresponding ones in the example above:
```
dimensions:
  climatological_time=3;
data:
  climatological_time=165, 196, 227;  // 1990-06-15, 1990-07-16, 1990-08-16
  climatological_time_bounds=151,10773, 181,10804, 212,10835;
    // 1990-06-01,2019-07-01, 1990-07-01,2019-08-01, 1990-08-01,2020-09-01
  month_indices=0, 1, 2, 0, 1, 2;
```

ch07.adoc
JonathanGregory: Thanks for your comments. I have put in a compromise version which uses some words and idea from each of you:

CF offers two conventions for describing anomaly data. In the remainder of this section and the following two (`<<anomalies-norm-data>>` and `<<anomalies-norm-metadata>>`), we describe a general convention that can describe anomalies over any type of coordinate, including details about how the norm was calculated. In `<<temporal-anomalies>>`, we describe a legacy convention that depends on special standard names. It can be used only for simple temporal anomalies, and is insufficiently informative for many use-cases.

Is that OK?

ch07.adoc
JonathanGregory: I have split the paragraph into two. The second paragraph becomes:

> This convention, using **`_anomaly`** standard names and **`reference_epoch`**, is considered to be a legacy. It is restricted to those common cases that have such standard names defined.

ch07.adoc
JonathanGregory: Sorry I made it unclear. It seemed to me that the text didn't explain what these "other dimensions" are. How about this, where the bold text is new:

> The generalized definition of an anomaly value _A_ of some physical quantity _q_ is the difference _P_ &minus; _N_ between a particular value _P_ of _q_ and a normal value or norm _N_ of _q_. _N_ is some statistic calculated from the values of _q_ that lie within specified ranges of one or more of **the variables (usually spatiotemporal coordinates) on which _q_ depends. We denote this set of variables as {_c_}; usually there is only one variable in the set.** _P_ can be, but is not necessarily, one of the set of values from which _N_ is calculated. 

> In the same way, a data variable _A_ containing anomalies (an "anomaly data variable") is notionally the difference between a data variable _P_ containing the original data and a data variable _N_ (a "norm data variable") containing the statistical norm. _P_ is typically not present in the dataset, and _N_ is usually absent as well.

> _P_ has all the same netCDF dimensions and coordinate variables as _A_. **_N_ shares all of them except for {_c_}. Instead of _A_'s dimensions and coordinate variables for {_c_}, _N_ has its own dimensions and coordinate variables for {_c_}. These dimensions of _N_ are the ones over which the norm is calculated.**

> The commonest kind of anomaly _A_ is a temporal anomaly, **where {_c_} is time**. "Temporal anomaly" means ...

ch07.adoc
JonathanGregory: Done, thanks

ch07.adoc
JonathanGregory: OK, changed.

ch07.adoc
JonathanGregory: Thanks, fixed.

ch07.adoc
JonathanGregory: Yes, quite right. I got trigger-happy with italics.

ch07.adoc
JonathanGregory: Quite so.

ch07.adoc
JonathanGregory: I agree. I have checked (maybe you did too) and there are no other offending italics.

ch07.adoc
JonathanGregory: Thanks

ch07.adoc
JonathanGregory: Ah yes, I see. I have made the second `delta_tas` example the same as the first, with a scalar `climatological_time`, while in the zonal-mean example I have added `zmlongitude` to the dimensions of `rtoa`.

ch07.adoc
JonathanGregory: You are quite right, thanks. I forgot to update the examples. Done now. Also, I noticed mistakes in the zonal-mean example, and have corrected it.

ch07.adoc
sethmcg: That works!  Much easier to follow.  Thanks!

JonathanGregory: Good. Thanks for checking. It's included now.

ch07.adoc
```diff
@@ -723,14 +725,559 @@ data:  // time coordinates translated to datetime format
...
+  double climatological_time_bounds(climatological_time,two);
+data:
+  time=15, 45, 76, 381, 411, 442;
+    // 2023-06-16, 2023-07-16, 2023-08-16, 2024-06-16, 2024-07-16, 2024-08-16
+  time_bounds=0,30, 30,61, 61,92, 366,396, 396,427, 427,458;
+    // beginning and end of Jun, Jul and Aug of 2023 and 2024
+  climatological_time=15, 45, ... 349;  // 1990-01-16, 1990-02-15 ... 1990-12-16
+  climatological_time_bounds=0,10623, 31,10651, ... 334,10957;
+    // 1990-01-01,2019-02-01, 1990-02-01,2019-03-01 ... 1990-12-01,2020-01-01
+  month_indices=5, 6, 7, 5, 6, 7;
```
davidhassell, line 1151: I'm somewhat lost with this example.
- I presume that `climatological_tas` should be `climatological_tas_metadata`, as per the caption
- The compression by gathering applies to the _data variable_, not the `climatological_time` coordinate variable
- The compressed variable's dimension is also the name of a coordinate variable (`time`), which causes problems when the variable is uncompressed to size 12

ch07.adoc
```diff
@@ -723,14 +725,559 @@ data:  // time coordinates translated to datetime format
...
+The value of element __i__ of the auxiliary coordinate variable is the index (numbering from 0) along the climatological time dimension of the norm corresponding to element __i__ of the anomaly time dimension.
+The norm metadata variable also has the same time dimension as the anomaly data variable (rather than being a scalar), but as usual, its values are arbitrary and meaningless.
+
+The climatological time axis of __N__ (i.e. its dimension and coordinate variable) must be included in the file, although __N__ itself need not be present.
+The auxiliary coordinate variable has a **`compress`** attribute naming the climatological time dimension, in order to make the link between them.
+This method of mapping between axes is equivalent to <<compression-by-gathering>>.
+Using this method means that the norm metadata variable can refer to a subset of elements of the climatological time axis if only some of them are relevant, and it can refer repeatedly to elements of climatological time axis if there is more than one anomaly time referring to a given climatological time.
```
davidhassell, line 1091: I don't think that this is how compression by gathering works in chapter 8 works. Compression by gathering applies to the data of the variable that carries the compressed dimension, and also references the index variable. This seems to a brand new mechanism that allows you to cherry-pick elements from an existing _coordinates_.

**Edit (18:18Z):** I have suggested an alternative, that just entails not using "compress" (in favour of "select"), and spelling out how the selection occurs.

ch07.adoc
```diff
@@ -723,14 +725,559 @@ data:  // time coordinates translated to datetime format
...
+
+Except in the case where __N__ has a multivalued climatological time axis (e.g., a monthly climatology), the norm metadata variable has a dimension and coordinate variable of size one, or a scalar coordinate variable, for each anomaly axis (although usually there is only one).
+Each such norm coordinate variable must have the same **`standard_name`** as the corresponding anomaly coordinate variable.
+Scalar norm coordinate variables must be named in the **`coordinates`** attribute of both the norm metadata variable and the anomaly data variable.
+Size-one dimensions of the norm metadata variable must also be dimensions of the anomaly data variable.
+The norm metadata variable must have no coordinate variables or scalar coordinate variables other than the norm coordinate variables, which are of size one.
+Therefore the norm metadata variable has only one element.
+It must have a **`_FillValue`** attribute, and its single element must be equal to the **`_FillValue`**, to indicate that it contains no meaningful data.
```
davidhassell: I don't think we should mandata a `_FillValue` attribute. The test is that it contains only missing values, and shouldn't get into the mine field of how the missing values are encoded.

ch07.adoc
davidhassell: OK - I've absorbed some of the new rules on how this works, and it's not really compression by gathering ... but it looks like it!

appa.adoc
```diff
@@ -102,8 +102,8 @@ Identifies the roles of mesh topology and location index set variables (see <<ap
 | **`compress`**
 | S
 | C
-| <<compression-by-gathering>>, <<reduced-horizontal-grid>>
-| Records dimensions which have been compressed by gathering.
+| <<reduced-horizontal-grid>>, <<anomalies-norm-metadata>>, <<compression-by-gathering>>
```
davidhassell: As per the compress->select suggestion
```suggestion
| <<reduced-horizontal-grid>>, <<compression-by-gathering>>
```

ch07.adoc
```diff
@@ -723,14 +725,559 @@ data:  // time coordinates translated to datetime format
...
+
+The mapping between the anomaly time axis of the anomaly variable and the climatological time axis of the norm is recorded by an auxiliary coordinate variable of integer type named by the **`coordinates`** attribute of the norm metadata variable.
+This auxiliary coordinate variable is one-dimensional and has the same time dimension as the anomaly data variable.
+The value of element __i__ of the auxiliary coordinate variable is the index (numbering from 0) along the climatological time dimension of the norm corresponding to element __i__ of the anomaly time dimension.
+The norm metadata variable also has the same time dimension as the anomaly data variable (rather than being a scalar), but as usual, its values are arbitrary and meaningless.
+
+The climatological time axis of __N__ (i.e. its dimension and coordinate variable) must be included in the file, although __N__ itself need not be present.
+The auxiliary coordinate variable has a **`compress`** attribute naming the climatological time dimension, in order to make the link between them.
```
davidhassell:
```suggestion
The auxiliary coordinate variable has a **`select`** attribute naming the climatological time dimension, in order to make the link between them.
```

ch07.adoc
```diff
@@ -723,14 +725,559 @@ data:  // time coordinates translated to datetime format
...
+This method of mapping between axes is equivalent to <<compression-by-gathering>>.
```
davidhassell:
```suggestion
The auxiliary coordinate variable implies the existence of another auxiliary coordinate variable of the same size, not in the file, whose data values are selected from the climatological time coordinate variable according to the climatological time axis index positions given in the data array.
```

ch07.adoc
```diff
@@ -723,14 +725,559 @@ data:  // time coordinates translated to datetime format
...
+    delta_tas:units="degC";
+    delta_tas:units_metadata="temperature: difference";
+    delta_tas:cell_methods="time: maximum time: anomaly_wrt climatological_tas";
+    delta_tas:ancillary_variables="climatological_tas_metadata";
+  int climatological_tas(time); // norm metadata variable
+    climatological_tas:coordinates="month_indices";
+    climatological_tas:cell_methods="climatological_time: mean within years
+      climatological_time: mean over years";
+  int month_indices(time);
+    month_indices:compress="climatological_time";
```
davidhassell: As part of the compress->select suggestion:
```suggestion
    month_indices:select="climatological_time";
```

ch07.adoc
davidhassell: I've made suggestions in compress->select above, which negates most of this comment thread, but:

- I presume that `climatological_tas` should be `climatological_tas_metadata`, as per the caption

still stands, I think.

appa.adoc
davidhassell, line 385: As per the compress->select suggestion
```suggestion
| **`select`**
| S
| C
| <<anomalies-norm-metadata>>
| Identifies other dimension to which the values in this variable are indices, used for selection of climatological time coordinates for norm metadata variables.

| **`source`**
```

ch07.adoc
sethmcg: I support using "select" instead of "compress".  I see the original motivation for re-using the functionality (which is very similar), but I agree that it's better to use different names to distinguish that they have different meanings.

appa.adoc
JonathanGregory: Done.

ch07.adoc
JonathanGregory: OK. I have therefore removed the recommendation that it "not have any of the attributes of `<<attribute-appendix>>` other than **`cell_methods`**, **`coordinates`** and **`_FillValue`**" because it's too complicated to rephrase this including several attributes which might be relevant to missing data. I've left it as "[The norm metadata variables] does not need attributes describing the __norm__ quantity (standard name, units, etc.), because they must be the same as for the anomaly data variable," and "its single element must indicate missing data."

ch07.adoc
JonathanGregory: Not done, as discussed

ch07.adoc
JonathanGregory: OK. `select` is a good name for it. It is not quite the same as `compress` anyway, because it allows repeated indices.

ch07.adoc
JonathanGregory: I'm sorry, I'm too tired to understand this sentence, so I've done only the deletion. The function of the aux coord var is explained in the previous paragraph.

ch07.adoc
JonathanGregory: I have changed `compress` to `select` and removed the sentence comparing the mechanism to compression by gathering. The way it works is described in the previous paragraph, which I have reordered thus:

> The norm metadata variable for a multivalued climatological time axis has the time dimension of the anomaly data variable as its sole dimension, and is thus itself multivalued, but its values are arbitrary and meaningless. The mapping between the anomaly time axis of the anomaly variable and the climatological time axis of the norm is recorded by an auxiliary coordinate variable of integer type named by the **`coordinates`** attribute of the norm metadata variable. The auxiliary coordinate variable is one-dimensional and has the same time dimension as the anomaly data variable. The climatological time axis of _N_ (i.e. its dimension and coordinate variable) must be included in the file, although _N_ itself need not be present. The auxiliary coordinate variable has a **`select`** attribute naming the climatological time dimension, in order to make the link between them. The value of element _i_ of the auxiliary coordinate variable is the index (numbering from 0) along the climatological time dimension of the norm corresponding to element _i_ of the anomaly time dimension. Using this method means that the norm metadata variable can refer to a subset of elements of the climatological time axis if only some of them are relevant, and it can refer repeatedly to elements of climatological time axis if there is more than one anomaly time referring to a given climatological time.

ch07.adoc
JonathanGregory: Yes, fine, thanks. Also I have corrected `climatological_tas_metadata`. Thanks for noticing.

appa.adoc
JonathanGregory: Also inserted a new entry in the table:

```
| **`select`**
| S
| C
| <<anomalies-norm-metadata>>
| Identifies a dimension to which the values in this variable are indices.
```

ch07.adoc
JonathanGregory: Done

ch07.adoc
```diff
@@ -723,14 +725,555 @@ data:  // time coordinates translated to datetime format
...
+----
+dimensions:
+  time=6;
+  climatological_time=12;
+variables:
+  float delta_tas(time,latitude,longitude); // anomaly data variable
+    delta_tas:standard_name="air_temperature";
+    delta_tas:units="degC";
+    delta_tas:units_metadata="temperature: difference";
+    delta_tas:cell_methods="time: maximum time: anomaly_wrt climatological_tas";
```
davidhassell:
```suggestion
    delta_tas:cell_methods="time: maximum time: anomaly_wrt climatological_tas_metadata";
```

JonathanGregory: Thanks

appa.adoc
JonathanGregory: Done

PR #605, opened 2025-09-19
Healpix

d70-t:
See issue #XXX for discussion of these changes.

# Release checklist
- [ ] Conformance document up to date? -- not yet, see [this comment](https://github.com/cf-convention/cf-conventions/issues/433#issuecomment-3312443938)

# For maintainers
Tags are set at the conclusion of the annual meeting; until then, `main` always is a draft for the next version.

### DIFF
--- a/appf.adoc
+++ b/appf.adoc
@@ -497,7 +497,7 @@ Domain: +
 // The cleanest-looking, which sometimes works for asciidoctor-pdf:
      **`-90.0 \<= latitude_of_projection_origin \<= 90.0`**
 
-| **`level`** | N
+| **`refinement_level`** | N
 | DGGS numerical order of a discrete global grid in the tessellation sequence. See "4.37. refinement level" in <<OGC_DGGS>>.
 
 | **`longitude_of_central_meridian`** | N
--- a/appf.adoc
+++ b/appf.adoc
@@ -112,6 +112,21 @@ Note also that the values "x" and "y" are not case-sensitive.
 The use of `projection_x_coordinate` and `projection_y_coordinate` was deprecated in version 1.9 of the CF Conventions.
 The initial definition of this projection used these standard names to identify the projection coordinates even though their canonical units (meters) do not match those required for this projection (radians).
 
+[[healpix]]
+=== HEALPix
+
+----
+grid_mapping_name = healpix
+----
+
+__Map parameters:__::
+* **`level`**
+* **`healpix_order`** - This parameter is optional (default is `nested`)
+
+__Map coordinates:__::
+
+__Notes:__::
+
 [[lambert-azimuthal-equal-area]]
 === Lambert azimuthal equal area
 
@@ -477,6 +492,9 @@ Domain: +
 // The cleanest-looking, which sometimes works for asciidoctor-pdf:
      **`-90.0 \<= latitude_of_projection_origin \<= 90.0`**
 
+| **`level`** | N
+| DGGS numerical order of a discrete global grid in the tessellation sequence. See "4.37. refinement level" in <<OGC_DGGS>>.
+
 | **`longitude_of_central_meridian`** | N
 | The line of longitude (degrees_east) at the center of a map projection generally used as the basis for constructing the projection.
 Domain: +
@@ -495,6 +513,9 @@ Domain: + **`-180.0 \<= longitude_of_projection_origin < 180.0`**
 
 | **`north_pole_grid_longitude`** | N
 | Longitude (degrees) of the true north pole in the rotated grid.
+
+| **`healpix_order`** | S
+| Identifier of the HEALPix cell ordering, either `nested` or `ring`. See <<GHB05>>
 
 | **`perspective_point_height`** | N
 | Records the height, __in metres__, of the map projection perspective point above the ellipsoid (or sphere).
--- a/appf.adoc
+++ b/appf.adoc
@@ -120,16 +120,30 @@ grid_mapping_name = healpix
 ----
 
 __Map parameters:__::
-* **`refinement_level`** - This parameter defines zoom refinement level.  
-* **`healpix_order`** - This parameter is optional (default is `nested`)
+* **`indexing_scheme`** - This parameter is optional (default is `nested`)
+* **`refinement_level`** - This parameter defines the HEALPix refinement level (0 == coarsest, optional if `indexing_scheme` is "nuniq" or "zuniq")
 
 __Map coordinates:__::
-The coordinate system is defined by discrete spherical cells arranged on a HEALPix grid. Each cell is represented by its cell center coordinates (latitude and longitude). Cells are indexed using either the nested or ring ordering scheme.
-The grid resolution is controlled by the zoom refinement level parameter **`refinement_level`**, starting from 0.  See <<GHB05>> for the calculation of center coordinates and cell boundaries coordinates. 
+The map uses a discrete 1-dimensional "index" coordinate, which uses the `standard_name` `"healpix_index"`, and can additionally be identified through the `grid_mapping` attribute.
+The coordinate can be any kind of coordinate, apart from primary or auxiliary coordinates, especially a tie-point coordinate can be used to encode contiguous blocks in index space.
 
+The coordinate system is defined by discrete cells on a sphere, arranged on a HEALPix grid.
+Each cell is represented by its cell index, which, together with the `indexing_scheme` and possible `refinement_level`, determines center coordinates (latitude and longitude) as well as boundaries.
+See <<GHB05>> for the calculation of center coordinates and cell boundaries coordinates.
 
 __Notes:__::
-At a given zoom refinement level, the total number of cells corresponds to 12 \times 4^{\text{refinement\_level}}
+
+Cells are indexed using either the nested, ring, nuniq or zuniq indexing scheme, the indices are stored in the index coordinate, which can be identified through the standard name `healpix_index`, or the `grid_mapping` attribute.
+
+For "nested" and "ring" scheme, the grid resolution is controlled by the refinement level parameter **`refinement_level`**, for "nuniq" and "zuniq", the resolution of a cell is implied by the index.
+
+An index according to the `nuniq` indexing scheme can be computed from the `refinement_level` and a `nested` index as follows <<RH15>>:
+
+`index_nuniq = index_nested + 4 ^ (1 + refinement_level)`
+
+TODO: add computation for `zuniq`. This seems to depend on `DEPTH_MAX`, which seems to depend on the number of available bits. How do we (formally) define DEPTH_MAX?
+
+At a given refinement level, the total number of cells corresponds to 12 \times 4^{\text{refinement\_level}}
 
 
 [[lambert-azimuthal-equal-area]]
@@ -519,8 +533,8 @@ Domain: + **`-180.0 \<= longitude_of_projection_origin < 180.0`**
 | **`north_pole_grid_longitude`** | N
 | Longitude (degrees) of the true north pole in the rotated grid.
 
-| **`healpix_order`** | S
-| Identifier of the HEALPix cell ordering, either `nested` or `ring`. See <<GHB05>>
+| **`indexing_scheme`** | S
+| Identifier of the HEALPix cell ordering, either `nested`, `ring`, `nuniq` or `zuniq`. See <<GHB05>> TODO: we need referenes or definitions for `nuniq` and `zuniq`.
 
 | **`perspective_point_height`** | N
 | Records the height, __in metres__, of the map projection perspective point above the ellipsoid (or sphere).
--- a/appf.adoc
+++ b/appf.adoc
@@ -120,16 +120,17 @@ grid_mapping_name = healpix
 ----
 
 __Map parameters:__::
-* **`level`**
+* **`level`** - This parameter defines zoom level.  
 * **`healpix_order`** - This parameter is optional (default is `nested`)
 
 __Map coordinates:__::
 The coordinate system is defined by discrete spherical cells arranged on a HEALPix grid. Each cell is represented by its cell center coordinates (latitude and longitude). Cells are indexed using either the nested or ring ordering scheme.
-The grid resolution is controlled by the zoom_level, starting from 0. At a given zoom level, the total number of cells is calculated as:
-\text{number of cells} = 12 \times 4^{\text{zoom\_level}}  See <<GHB05>> for the calculation of center coordinates and cell boundaries coordinates. 
+The grid resolution is controlled by the zoom level parameter **`level`**, starting from 0.  See <<GHB05>> for the calculation of center coordinates and cell boundaries coordinates. 
 
 
 __Notes:__::
+At a given zoom level, the total number of cells corresponds to 12 \times 4^{\text{zoom\_level}}
+
 
 [[lambert-azimuthal-equal-area]]
 === Lambert azimuthal equal area
--- a/appf.adoc
+++ b/appf.adoc
@@ -124,6 +124,10 @@ __Map parameters:__::
 * **`healpix_order`** - This parameter is optional (default is `nested`)
 
 __Map coordinates:__::
+The coordinate system is defined by discrete spherical cells arranged on a HEALPix grid. Each cell is represented by its cell center coordinates (latitude and longitude). Cells are indexed using either the nested or ring ordering scheme.
+The grid resolution is controlled by the zoom_level, starting from 0. At a given zoom level, the total number of cells is calculated as:
+\text{number of cells} = 12 \times 4^{\text{zoom\_level}}  See <<GHB05>> for the calculation of center coordinates and cell boundaries coordinates. 
+
 
 __Notes:__::
 
--- a/appf.adoc
+++ b/appf.adoc
@@ -120,16 +120,16 @@ grid_mapping_name = healpix
 ----
 
 __Map parameters:__::
-* **`level`** - This parameter defines zoom level.  
+* **`refinement_level`** - This parameter defines zoom refinement level.  
 * **`healpix_order`** - This parameter is optional (default is `nested`)
 
 __Map coordinates:__::
 The coordinate system is defined by discrete spherical cells arranged on a HEALPix grid. Each cell is represented by its cell center coordinates (latitude and longitude). Cells are indexed using either the nested or ring ordering scheme.
-The grid resolution is controlled by the zoom level parameter **`level`**, starting from 0.  See <<GHB05>> for the calculation of center coordinates and cell boundaries coordinates. 
+The grid resolution is controlled by the zoom refinement level parameter **`refinement_level`**, starting from 0.  See <<GHB05>> for the calculation of center coordinates and cell boundaries coordinates. 
 
 
 __Notes:__::
-At a given zoom level, the total number of cells corresponds to 12 \times 4^{\text{zoom\_level}}
+At a given zoom refinement level, the total number of cells corresponds to 12 \times 4^{\text{refinement\_level}}
 
 
 [[lambert-azimuthal-equal-area]]
--- a/bibliography.adoc
+++ b/bibliography.adoc
@@ -20,6 +20,7 @@ Federal Geographic Data Committee, FGDC-STD-001-1998.
 - [[[OGC_DGGS]]]  link:$$http://www.opengis.net/doc/AS/dggs/2.0$$[Topic 21 - Discrete Global Grid Systems - Part 1 Core Reference system and Operations and Equal Area Earth Reference System].
 OGC document 12-063. 1st May 2015.
 - [[[OGP-EPSG_GN7_2]]]  link:$$https://epsg.org/guidance-notes.html$$[OGP Surveying and Positioning Guidance Note 7, part 2: Coordinate Conversions and Transformations including Formulas].
+- [[[RH15]]] link:$$https://doi.org/10.1051/0004-6361/201526549$$[Efficient data structures for masks on 2D grids]. M Reinecke and E Hivon. 2015. _Astronomy & Astrophysics_. 580, A132
 - [[[SCH02]]] link:$$https://doi.org/10.1175/1520-0493(2002)130<2459:ANTFVC>2.0.CO;2$$[A new terrain-following vertical coordinate formulation for atmospheric prediction models]. C Schaer, D Leuenberger, and O Fuhrer. 2002. _Monthly Weather Review_.  130. 2459-2480.
 - [[[Snyder]]]  link:$$https://doi.org/10.3133/pp1395$$[Map Projections: A Working Manual]. USGS Professional Paper 1395.
 - [[[UDUNITS]]]  link:$$https://doi.org/10.5065/D6KD1WN0$$[UDUNITS Software Package].  UNIDATA Program Center of the University Corporation for Atmospheric Research.
--- a/bibliography.adoc
+++ b/bibliography.adoc
@@ -10,12 +10,14 @@ Sponsored by the "Cooperative Ocean/Atmosphere Research Data Service," a NOAA/un
 - [[[DCG19]]]  link:$$https://doi.org/10.5194/gmd-12-4099-2019$$[Evaluation of lossless and lossy algorithms for the compression of scientific datasets in netCDF-4 or HDF5 files]. Delaunay, X., A. Courtois, and F. Gouillon: _Geosci. Model Dev._, 12, 4099-4113, 2019.
 - [[[FGDC]]]  link:$$https://www.fgdc.gov/standards/projects/FGDC-standards-projects/metadata/base-metadata/v2_0698.pdf$$[Content Standard for Digital Geospatial Metadata].
 Federal Geographic Data Committee, FGDC-STD-001-1998.
+- [[[GHB05]]]  link:$$https://doi.org/10.1086/427976$$[HEALPix: A Framework for High-Resolution Discretization and Fast Analysis of Data Distributed on the Sphere].
 - [[[IEEE_754]]]  link:$$https://doi.org/10.1109/IEEESTD.2019.8766229$$[IEEE Standard for Floating-Point Arithmetic], in _IEEE Std 754-2019 (Revision of IEEE 754-2008)_, 22 July 2019.
 - [[[Kou21]]]  link:$$https://doi.org/10.5194/gmd-14-377-2021$$[A note on precision-preserving compression of scientific data]. Kouznetsov, R.: _Geosci. Model Dev._, 14, 377-389, 2021.
 - [[[KRD21]]]  link:$$https://doi.org/10.1038/s43588-021-00156-2$$[Compressing atmospheric data into its real information content]. Klöwer, M., Razinger, M., Dominguez, J. J., Düben, P., and Palmer, T. N.: _Nat. Comput. Sci._, 1, 713-724, 2021.
 - [[[NetCDF]]]  link:$$https://doi.org/10.5065/D6H70CW6$$[NetCDF Software Package].  UNIDATA Program Center of the University Corporation for Atmospheric Research.
 - [[[NUG]]]  link:$$https://docs.unidata.ucar.edu/nug/current/index.html$$[The NetCDF User's Guide].
 - [[[OGC_WKT-CRS]]]  link:$$https://www.opengeospatial.org/standards/wkt-crs$$[OGC Well-known text representation of coordinate reference systems].
+- [[[OGC_DGGS]]]  link:$$http://www.opengis.net/doc/AS/dggs/2.0$$[Topic 21 - Discrete Global Grid Systems - Part 1 Core Reference system and Operations and Equal Area Earth Reference System].
 OGC document 12-063. 1st May 2015.
 - [[[OGP-EPSG_GN7_2]]]  link:$$https://epsg.org/guidance-notes.html$$[OGP Surveying and Positioning Guidance Note 7, part 2: Coordinate Conversions and Transformations including Formulas].
 - [[[SCH02]]] link:$$https://doi.org/10.1175/1520-0493(2002)130<2459:ANTFVC>2.0.CO;2$$[A new terrain-following vertical coordinate formulation for atmospheric prediction models]. C Schaer, D Leuenberger, and O Fuhrer. 2002. _Monthly Weather Review_.  130. 2459-2480.
--- a/history.adoc
+++ b/history.adoc
@@ -12,6 +12,7 @@
 * {issues}583[Issue #583]: Correct "most rapidly varying dimension" in terminology section.
 * {issues}584[Issue #584]: Allow `Z` as time-zone offset, with a couple of examples, and allow time-zone offset with date alone, both being consistent with UDUNITS syntax.
 * {issues}508[Issue #508]: Introduce aggregation variables
+* {issues}433[Issue #433]: Convention for HEALPix grid parameters
 
 === Version 1.12 (04 December 2024)
 
--- a/ch05.adoc
+++ b/ch05.adoc
@@ -428,37 +428,6 @@ dimensions:
 ----
 ====
 
-[[healpix-with-compression-by-gathering]]
-[caption="Example 5.11. "]
-.Healpix with compression by gathering
-====
-----
-dimensions:
-    time = 100;
-    // TODO: the cell dimension won't be available when encoding in zarr (as no variable uses that dimension)
-    // TODO: this might or might not be a problem (the size is always 12*4**crs:refinement_level)
-    cell = 49152;
-    cell_id = 15234;
-  variables:
-    float time(time) ;
-      time.units = "days since 2024-09-04" ;
-    float temp(time, cell_id) ;
-      temp:standard_name = "air_temperature" ;
-      temp:units = "K" ;
-      temp:grid_mapping = "crs: cell" ;
-    float pres(time, cell_id) ;
-      pres:standard_name = "air_pressure" ;
-      pres:units = "Pa" ;
-      temp:grid_mapping = "crs: cell" ;
-    int cell_id(cell_id) ;
-      cell_id:compress = "cell" ;
-    int crs ;
-	  crs:grid_mapping_name = "healpix" ;
-      crs:refinement_level = 7 ;
-      crs:healpix_order = "nested" ;
-----
-====
-
 [[use-of-the-crs-well-known-text-format, Section 5.6.1, "Use of the CRS Well-known Text Format"]]
 ==== Use of the CRS Well-known Text Format
 
--- a/ch05.adoc
+++ b/ch05.adoc
@@ -428,6 +428,37 @@ dimensions:
 ----
 ====
 
+[[healpix-with-compression-by-gathering]]
+[caption="Example 5.11. "]
+.Healpix with compression by gathering
+====
+----
+dimensions:
+    time = 100;
+    // TODO: the cell dimension won't be available when encoding in zarr (as no variable uses that dimension)
+    // TODO: this might or might not be a problem (the size is always 12*4**crs:level)
+    cell = 49152;
+    cell_id = 15234;
+  variables:
+    float time(time) ;
+      time.units = "days since 2024-09-04" ;
+    float temp(time, cell_id) ;
+      temp:standard_name = "air_temperature" ;
+      temp:units = "K" ;
+      temp:grid_mapping = "crs: cell" ;
+    float pres(time, cell_id) ;
+      pres:standard_name = "air_pressure" ;
+      pres:units = "Pa" ;
+      temp:grid_mapping = "crs: cell" ;
+    int cell_id(cell_id) ;
+      cell_id:compress = "cell" ;
+    int crs ;
+	  crs:grid_mapping_name = "healpix" ;
+      crs:level = 7 ;
+      crs:healpix_order = "nested" ;
+----
+====
+
 [[use-of-the-crs-well-known-text-format, Section 5.6.1, "Use of the CRS Well-known Text Format"]]
 ==== Use of the CRS Well-known Text Format
 
--- a/ch05.adoc
+++ b/ch05.adoc
@@ -436,7 +436,7 @@ dimensions:
 dimensions:
     time = 100;
     // TODO: the cell dimension won't be available when encoding in zarr (as no variable uses that dimension)
-    // TODO: this might or might not be a problem (the size is always 12*4**crs:level)
+    // TODO: this might or might not be a problem (the size is always 12*4**crs:refinement_level)
     cell = 49152;
     cell_id = 15234;
   variables:
@@ -454,7 +454,7 @@ dimensions:
       cell_id:compress = "cell" ;
     int crs ;
 	  crs:grid_mapping_name = "healpix" ;
-      crs:level = 7 ;
+      crs:refinement_level = 7 ;
       crs:healpix_order = "nested" ;
 ----
 ====


appf.adoc
```diff
@@ -112,6 +112,40 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+
+Cells are indexed using either the nested, ring, nuniq or zuniq indexing scheme, the indices are stored in the index coordinate, which can be identified through the standard name `healpix_index`, or the `grid_mapping` attribute.
+
+For "nested" and "ring" scheme, the grid resolution is controlled by the refinement level parameter **`refinement_level`**, for "nuniq" and "zuniq", the resolution of a cell is implied by the index.
+
+An index according to the `nuniq` indexing scheme can be computed from the `refinement_level` and a `nested` index as follows <<RH15>>:
+
+`index_nuniq = index_nested + 4 ^ (1 + refinement_level)`
+
+TODO: add computation for `zuniq`. This seems to depend on `DEPTH_MAX`, which seems to depend on the number of available bits. How do we (formally) define DEPTH_MAX?
```
carloshorn: A bit lengthy, but maybe something like the following:

The `zuniq` indexing represents a a fixed-point arithmetic real number `z in [0, 12)`. The integer part of `z` encodes the HEALPix face and the fractional part encodes the position along the space-filling curve within the face. This provides a surjective map from `[0, 12)` to the unit sphere S2.
The precision of that real number is given by the available bits, where the integer part takes 4 bits and the fractional part requires two bits per level followed by a sentinel bit which sets the location along the space-filling curve to the cell center. This allows for a maximum level of `L_max = (64 - 4 - 1) // 2 = 29` for 64 available bits. The `zuniq` index can be computed from the `refinement_level`, the `nested` index and a given `L_max` as follows:

`index_zuniq = (2 * index_nested + 1) * 4 ^ (L_max - refinement_level)`

Is it maybe too hard to follow? I could also add an example.

carloshorn: Note on different bit length: when using 32-bit integers, `L_max = 13`. When converting to 64-bit integers, we have to set the leading bits, i.e. `zuniq_64 = zuniq_32 << 32` or in other words, we have to pre- or append four zero bytes depending on little- or big endianness. I guess this information is more confusing than helpful, so I would just recommend using 64-bit integers.
Maybe also worth mentioning is the fact that we do not use the sign bit, so it does not matter if we use signed or unsigned integers for the representation, so I would recommend signed 64-bit integers.

d70-t: My impression would be, that having two formulas isn't that bad, I think something like:

```
L_max = (n_bits - 4 - 1) // 2
index_zuniq = (2 * index_nested + 1) * 4 ^ (L_max - refinement_level)
```

is reasonably concise, and it's probably not necessary to specify all the clever mathematical details for this specification document and would outsource this to external references (e.g. the one you suggested at https://github.com/cf-convention/cf-conventions/issues/433#issuecomment-3315946477).

The thing which troubles me the most, is where to specify `n_bits`.
* Do we have another metadata field?
* Do we state, that this is always the number of bits of the integer data type used?
* How does this play with e.g. Python data types (where numbers are of arbitrary size)?
* How does this play with e.g. tie-point coordinates, where technically, the cell index values are interpolated, and the data type of the interpolated variable is (at least in theory) dependent on the client-side implementation.

carloshorn: > Do we have another metadata field?

I would say we do not need another field. I do not see any reason to encode `zuniq` in a weird custom integer type. Only if somebody thinks that HEALPix is still useful below cm resolution and therefore needs more than 64-bits which I doubt.

> Do we state that this is always the number of bits of the integer data type used?

I would say yes and connect the number of bits with the integer data type used within the netCDF file and recommend 64-bit integers for non-expert users. Even if we only store cells at a level below 13 which would fit into a 32-bit integer, using the shuffle filter with compression would shrink these zeros away, so there is not even an argument on saving space, but I do not want to take away the possibility to use only 32-bit for expert users who want to load and directly stream data in the most optimized way.

> How does this play with e.g. Python data types (where numbers are of arbitrary size)?

I would delegate the responsibility of keeping track of data types to the applications. Most users in Python will likely use libraries like NumPy to handle such data, so they will likely be in a typed ecosystem anyway.

> How does this play with e.g. tie-point coordinates, where technically, the cell index values are interpolated, and the data type of the interpolated variable is (at least in theory) dependent on the client-side implementation.

The interpolation has the same limitation as for the nested index that if it is floating point based, it will be limited by the number of significand bits. An IEEE 754 double-precision float has 52 significand bits. Therefore, it can only interpolate down to `Lmax = (52-4-1)//2 = 23`. In case of `zuniq`, the trailing zeros are picked up by the exponent.
However, this is likely more a problem with `nested` indexing, because `zuniq` is more often used for MOCs which combine consecutive indices to lower-level cells already.
Edit: regarding the data type in an interpolation case, the index tie point variable will still have the integer type we want to cast the result to, so I do not see any ambiguity here. The interpolation variable takes the attribute `computational_precision` which when set to 64 indicates the floating-point type, so the Unkown might be the number of significand bits where I would expect something in the order of 50, so below level 20 should be on the safe side. On the other hand, double precision in netCDF is always IEEE 754, so maybe assuming 52 significant bits is reasonable.

I think, we may somehow put a warning sign about the tie-point interpolation somewhere. In the worst case, delta compression can still help avoid wasting storage for the index.

carloshorn: We may also go with the s2geometry approach and define the `zuniq` index with 64-bit integers. An expert user is not forced to follow cf-conventions... What do you think?

This choice is supported by existing libraries, e.g. [healpy](https://github.com/healpy/healpy/blob/773c23b92939494c3564692c6ba756e24c73f66a/lib/healpy/pixelfunc.py#L110) and [cdshealpix](https://github.com/cds-astro/cds-healpix-rust/blob/48a5ee396abf852450e4b685f6e107310f39beec/src/lib.rs#L61) define the maximum level as a constant set to 29.

appf.adoc
```diff
@@ -112,6 +112,40 @@ Note also that the values "x" and "y" are not case-sensitive.
...
 
+[[healpix]]
+=== HEALPix
+
+----
+grid_mapping_name = healpix
+----
+
+__Map parameters:__::
+* **`indexing_scheme`** - This parameter is optional (default is `nested`)
```
davidhassell: I think `indexing_scheme` should be mandatory.

appf.adoc
```diff
@@ -112,6 +112,40 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+* **`refinement_level`** - This parameter defines the HEALPix refinement level (0 == coarsest, optional if `indexing_scheme` is "nuniq" or "zuniq")
```
davidhassell: Must be omitted for `nuniq` and `zuniq`.

appf.adoc
```diff
@@ -477,6 +525,9 @@ Domain: +
 // The cleanest-looking, which sometimes works for asciidoctor-pdf:
      **`-90.0 \<= latitude_of_projection_origin \<= 90.0`**
 
+| **`refinement_level`** | N
+| DGGS numerical order of a discrete global grid in the tessellation sequence. See "4.37. refinement level" in <<OGC_DGGS>>.
```
carloshorn: After the conversation about relating DGGRS -> DGGS -> DGGH, I am not sure about using an abbreviation in this text. Can we just call it the level within the cell hierarchy while keeping the reference for providing details in the DGG(R)S context?

davidhassell: Might it be better just to define this explicitly, and in relation to HEALPix:
```suggestion
| The refinement level of the grid within the HEALPix hierarchy, starting at 0 for the base tessellation with 12 cells globally, 1 for the next level with 48 cells globally, etc.
```
We've saved ourselves a lot of pain by _not_ being completely general along the lines of DGG*, so no need to start now :). If in the future we wanted to general the description of `refinement_level` so that it could _also_  be used with other types of grid, then that will not be a problem.

appf.adoc
```diff
@@ -112,6 +112,54 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+----
+grid_mapping_name = healpix
+----
+
+__Map parameters:__::
+* **`indexing_scheme`**
+* **`refinement_level`** - Mandatory for `indexing_scheme` "nested" or "ring", must be omitted for "nuniq" or "zuniq".
+
+__Map coordinates:__::
+The map uses a discrete 1-dimensional "index" coordinate, which uses the `standard_name` `"healpix_index"`, and can additionally be identified through the `grid_mapping` attribute.
```
davidhassell:
```suggestion
The integer-valued HEALPix index coordinates are identified by the **`standard_name`** attribute value **`healpix_index`**
```
We don't mention the extended grid mapping syntax for other grid mappings, so no need to here, I think. Re-worded to be consistent with the other grid mapping descriptions.

appf.adoc
```diff
@@ -112,6 +112,54 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+The coordinate can be any kind of coordinate, apart from coordinate variables or auxiliary coordinates, especially a tie-point coordinate can be used to encode contiguous blocks in index space in a compact way.
```
davidhassell:
```suggestion
The HEALPix index coordinates may be stored in a coordinate, auxiliary, or scalar coordinate variable. Indexing schemes are optimized to allow applications efficient data access when the HEALPix indices are ordered strictly monotonically. Therefore it is recommended to store strictly monotonic HEALPix indices in a coordinate variable, which allows an application to assess their monotonicity without expensive checking. Note that it may be possible to encode consecutive HEALPix indices in a compact way with a <<compression-by-coordinate-subsampling, tie point coordinate variable>>.
```

appf.adoc
```diff
@@ -112,6 +112,54 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+If tie-point coordinates are used, take special care of the required precision.
```
davidhassell:
```suggestion
```
Not sure we need this, as no extra care is needed over normal coordinates (i.e. the tie point variable has to have the data type of the uncompressed coordinates. 

That said, we do need a small addition to chapter 8 that says something along the lines of "the interpolated tie points should be cast to the data type of the tie points". This is best done in a separate issue (are there any impacts on this statement to other use cases?... in which case that's another reason to not mention this here and now).

appf.adoc
```diff
@@ -112,6 +112,54 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+
++
+The coordinate system is defined by discrete cells on a sphere, arranged on a HEALPix grid.
+Each cell is represented by its cell index, which, together with the `indexing_scheme` and possible `refinement_level`, determines center coordinates (latitude and longitude) as well as boundaries.
+See <<GHB05>> for the calculation of center coordinates and cell boundary coordinates.
+
+__Notes:__::
+
+Cells are indexed using either the nested, ring, nuniq or zuniq indexing scheme, the indices are stored in the index coordinate, which can be identified through the standard name `healpix_index`, or the `grid_mapping` attribute.
```
davidhassell:
```suggestion
Cells are indexed using either the nested, ring, nuniq or zuniq indexing scheme.
```
No need to repeat the bit in "Map coordinates"

appf.adoc
```diff
@@ -112,6 +112,54 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+
++
+An index according to the `nuniq` indexing scheme can be computed from the `refinement_level` and a `nested` index as follows <<RH15>>:
+
++
+`index_nuniq = index_nested + 4 ^ (1 + refinement_level)`
+
++
+The `zuniq` indexing scheme depends on the number of bits used by the integer type.
+This specification assumes 64 bit integers for all `zuniq` indices.
```
davidhassell:
```suggestion
```
We already restrict refinement_level <= 29, so I don't see the advantage in talking about bits and precision, here.

bibliography.adoc
```diff
@@ -10,14 +10,18 @@ Sponsored by the "Cooperative Ocean/Atmosphere Research Data Service," a NOAA/un
...
 - [[[Kou21]]]  link:$$https://doi.org/10.5194/gmd-14-377-2021$$[A note on precision-preserving compression of scientific data]. Kouznetsov, R.: _Geosci. Model Dev._, 14, 377-389, 2021.
 - [[[KRD21]]]  link:$$https://doi.org/10.1038/s43588-021-00156-2$$[Compressing atmospheric data into its real information content]. Klöwer, M., Razinger, M., Dominguez, J. J., Düben, P., and Palmer, T. N.: _Nat. Comput. Sci._, 1, 713-724, 2021.
+- [[[K24]]] link:$$https://arxiv.org/html/2501.04705v1$$[Space-Filling Curves]. Kanungo, S.: _arXiv.org_, 2024
 - [[[NetCDF]]]  link:$$https://doi.org/10.5065/D6H70CW6$$[NetCDF Software Package].  UNIDATA Program Center of the University Corporation for Atmospheric Research.
 - [[[NUG]]]  link:$$https://docs.unidata.ucar.edu/nug/current/index.html$$[The NetCDF User's Guide].
 - [[[OGC_WKT-CRS]]]  link:$$https://www.opengeospatial.org/standards/wkt-crs$$[OGC Well-known text representation of coordinate reference systems].
+- [[[OGC_DGGS]]]  link:$$http://www.opengis.net/doc/AS/dggs/2.0$$[Topic 21 - Discrete Global Grid Systems - Part 1 Core Reference system and Operations and Equal Area Earth Reference System].
```
davidhassell:
```suggestion
```

bibliography.adoc
```diff
@@ -10,14 +10,18 @@ Sponsored by the "Cooperative Ocean/Atmosphere Research Data Service," a NOAA/un
 - [[[DCG19]]]  link:$$https://doi.org/10.5194/gmd-12-4099-2019$$[Evaluation of lossless and lossy algorithms for the compression of scientific datasets in netCDF-4 or HDF5 files]. Delaunay, X., A. Courtois, and F. Gouillon: _Geosci. Model Dev._, 12, 4099-4113, 2019.
 - [[[FGDC]]]  link:$$https://www.fgdc.gov/standards/projects/FGDC-standards-projects/metadata/base-metadata/v2_0698.pdf$$[Content Standard for Digital Geospatial Metadata].
 Federal Geographic Data Committee, FGDC-STD-001-1998.
+- [[[GHB05]]]  link:$$https://doi.org/10.1086/427976$$[HEALPix: A Framework for High-Resolution Discretization and Fast Analysis of Data Distributed on the Sphere].
```
davidhassell:
```suggestion
- [[[GHB05]]]  link:$$https://doi.org/10.1086/427976$$[HEALPix: A Framework for High-Resolution Discretization and Fast Analysis of Data Distributed on the Sphere]. K. Gorski, Eric Hivon, A. Banday, B. Wandelt, M. Bartelmann, et al.: The Astrophysical Journal, 2005, 622 (2), pp.759-771.
```

appf.adoc
```diff
@@ -112,6 +112,54 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+The coordinate system is defined by discrete cells on a sphere, arranged on a HEALPix grid.
+Each cell is represented by its cell index, which, together with the `indexing_scheme` and possible `refinement_level`, determines center coordinates (latitude and longitude) as well as boundaries.
+See <<GHB05>> for the calculation of center coordinates and cell boundary coordinates.
+
+__Notes:__::
+
+Cells are indexed using either the nested, ring, nuniq or zuniq indexing scheme, the indices are stored in the index coordinate, which can be identified through the standard name `healpix_index`, or the `grid_mapping` attribute.
+
++
+For "nested" and "ring" scheme, the grid resolution is controlled by the refinement level parameter `refinement_level`.
```
davidhassell:
```suggestion
For "nested" and "ring" scheme, the grid resolution is controlled by the refinement level parameter **`refinement_level`** <<GHB05>>.
```

appf.adoc
```diff
@@ -496,6 +547,9 @@ Domain: + **`-180.0 \<= longitude_of_projection_origin < 180.0`**
 | **`north_pole_grid_longitude`** | N
 | Longitude (degrees) of the true north pole in the rotated grid.
 
+| **`indexing_scheme`** | S
+| Identifier of the HEALPix cell indexing scheme, either `nested`, `ring`, `nuniq` or `zuniq`. See <<healpix>> for more detailed definitions.
```
davidhassell:
```suggestion
| The HEALPix cell indexing scheme, either **`nested`**, **`ring`**, **`nuniq`** or **`zuniq`**. See <<healpix>> for more detailed definitions.
```

appf.adoc
```diff
@@ -112,6 +112,54 @@ Note also that the values "x" and "y" are not case-sensitive.
...
++
+The `zuniq` indexing scheme depends on the number of bits used by the integer type.
+This specification assumes 64 bit integers for all `zuniq` indices.
+An index according to the `zuniq` indexing scheme can be computed from the `refinement_level` and a `nested` index as follows <<K24>>:
+
++
+`index_zuniq = (2 * index_nested + 1) * 4 ^ (29 - refinement_level)`
+
++
+The value `29` is the maximum representable `refinement_level` and is computed from the number of bits as `(n_bits - 4 - 1) // 2`.
```
davidhassell:
```suggestion
The value `29` is the maximum representable `refinement_level` (which is the highest level for which all indices can be represented by 64-bit integers).
```
What about `unsigned 64-bit`? Are we OK with that ... CF does not allow anything larger.

carloshorn: See https://github.com/cf-convention/cf-conventions/pull/605#discussion_r2369595152, in short, it does not matter if we encode `zuniq` in signed or unsigned integers, because it has an odd number of bits.
I personally am okay with 64 bits, because we are talking about cm resolution, which is more than enough IMHO.

appf.adoc
d70-t: Probably this statement is too much, I agree. However, there might be some additional issue, because [tie point coordinate interpolation only specifies 32 or 64 as `computational_precision`](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.12/cf-conventions.html#compression-by-coordinate-subsampling-interpolation-method-implementation), and 64 means 64bit float, which is only 52 bit integer. Whereas a 64 bit integer coordinate variable would actually be able to correctly represent 12 more bits (and thus 6 more refinement levels).

So even if the interpolated coordinate value is cast to 64 bit int after interpolation, the value might be wrong, if the interpolation was carried out as double. Options to solve that would be:

* recommend using coordinate variables instead (may cause memory issues though)
* don't support finer `refinement_levels` (not so nice)
* extend the definition of tie point `computational_precision` to more than 64bit float (certainly needs a new issue)

To be clear on this particular thread: I'd be fine dropping (or extending) this sentence, but I wanted to check back with you.

appf.adoc
```diff
@@ -496,6 +546,9 @@ Domain: + **`-180.0 \<= longitude_of_projection_origin < 180.0`**
 | **`north_pole_grid_longitude`** | N
 | Longitude (degrees) of the true north pole in the rotated grid.
 
+| **`indexing_scheme`** | S
```
davidhassell: Can we move this to its alphabetical position in the table?

appf.adoc
```diff
@@ -112,6 +112,53 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+For the "nuniq" and "zuniq" schemes, the resolution of a cell is implied by the index, and `refinement_level` must be omitted.
+
++
+An index according to the `nuniq` indexing scheme can be computed from the `refinement_level` and a `nested` index as follows <<RH15>>:
+
++
+`index_nuniq = index_nested + 4 ^ (1 + refinement_level)`
+
++
+The `zuniq` indexing scheme depends on the number of bits used by the integer type.
```
davidhassell: I'm still a bit confused about bits! We say "zuniq` indexing scheme depends on the number of bits used by the integer type" and then proceed to state it's definition in term of decimal arithmetic, irrespective of the data type. I appreciate that the formula comes about from a consideration of bits, but I don't think we need to mention them here?

It might be good to, though, to mention the difference between nuniq and zuniq (i.e. one is optimised for access within a refinement level, and the other optimised for access across refinement levels - if I've understood correctly).

Similarly maybe we could discern ring and nested as optimised for zonal access and <whatever-it's-called> access?

appf.adoc
```diff
@@ -112,6 +112,53 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+----
+grid_mapping_name = healpix
+----
+
+__Map parameters:__::
+* **`indexing_scheme`**
+* **`refinement_level`** - Mandatory for `indexing_scheme` "nested" or "ring", must be omitted for "nuniq" or "zuniq".
+
+__Map coordinates:__::
+The integer-valued HEALPix index coordinates are identified by the **`standard_name`** attribute value **`healpix_index`**
```
davidhassell:
```suggestion
The integer-valued HEALPix index coordinates are identified by the **`standard_name`** attribute value **`healpix_index`**.
```

appf.adoc
davidhassell: Great points. My current thinking is to leave this recommendation out completely for now (i.e. delete the whole of "_Note that it may be possible to encode consecutive HEALPix indices in a compact way with a <<compression-by-coordinate-subsampling, tie point coordinate variable>>. If tie-point coordinates are used, take special care of the required precision_".

There are just too many fiddly things here, not least that tie-point interpolation currently does not contain any guidance on how to deal with the integer case. Better to work out how to make sure that integer coordinates can be faithfully reconstructed from tie points, in general, and then we can put back in the guidance sentence here.

I'm not sure off hand how that would work - but a new interpolation method of `linear_integer` might be the way to go. Whatever the solution, it'll be easy to implement, but not on the timescales for getting this into CF-1.13 - so we can take it up as another issue and sort it out independently.

appf.adoc
```diff
@@ -112,6 +112,53 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+
++
+The `zuniq` indexing scheme depends on the number of bits used by the integer type.
+An index according to the `zuniq` indexing scheme can be computed from the `refinement_level` and a `nested` index as follows <<K24>>:
+
++
+`index_zuniq = (2 * index_nested + 1) * 4 ^ (29 - refinement_level)`
+
++
+The value `29` is the maximum representable `refinement_level` (which is the highest level for which all indices can be represented by 64-bit integers).
```
davidhassell:
```suggestion
The value `29` is the maximum allowed `refinement_level` (which is the highest level for which all indices can be represented by 64-bit integers).
```

davidhassell (commented):
A few more comments - homing in!

appf.adoc
carloshorn: > [...] <whatever-it's-called> access

I would call it iso-latitude ring access. Even though I would prefer a formulation which just describes the ordering of pixels instead of the access pattern which may work best. Something like:

While ring indexing sorts pixels along iso-latitude rings from north to south, nested indexing sorts pixels along the Z-order curve, resulting in spatial proximity for consecutive pixels.

While the nuniq indexing concatenates the nested indexing from the lower to the higher level, the zuniq indexing sorts pixels of any level along the Z-order curve.

appf.adoc
d70-t: I agree. While I still think, tie point coordinates are super helpful, we probably shouldn't advertise them actively before solving the little details. And removing the comment doesn't forbid to use it :-)

appf.adoc
davidhassell: I've had a go at incorporating your text, @carloshorn, which rearranged things a bit. The content's basically the same as before, with Carlos's text, and I've added a note about cell `0` being at 45 degrees east.

Here's the rendered text, with the raw asciidoc below:

----

Cells are indexed using either the **`nested`**, **`ring`**, **`nuniq`** or **`zuniq`** indexing scheme.

The **`ring`** scheme sorts pixels along iso-latitude rings from north to south; and **`nested`** indexing sorts pixels along the Z-order curve, resulting in spatial proximity for consecutively pixels.

For both the **`nested`** and **`ring`** schemes, the grid resolution is controlled by the refinement level parameter **`refinement_level`** <<GHB05>>.
The total number of cells on the sphere is `12 * (4 ^ refinement_level)`.
The lowest allowed refinement level is `0`, giving 12 cells globally.
The algorithm that assigns cell locations to **`nested`** or **`ring`** indices at a given refinement level is defined in <<GHB05>>. In this algorithm, a cell with index `0`, at any refinment level, always has its southern vertex on the meridian at 45 degrees east.

The **`nuniq`** and **`zuniq`** indexing schemes allow pixels from multiple refinement levels to be combined: **`nuniq`** effectively concatenates **`nested`** indices from lower to the higher refinement levels, resulting in spatial proximity for consecutive pixels within a refinement level but not across different refinement levels; and **`zuniq`** indexing sorts pixels of any refinement level along the Z-order curve, resulting in spatial proximity for consecutive pixels, regardless of their refinement level.

An index according to the **`nuniq`** indexing scheme can be computed from a refinement level and a **`nested`** index with the formula `index_nuniq = index_nested + 4 ^ (1 + refinement_level)` <<RH15>>.

An index according to the **`zuniq`** indexing scheme can be computed from a refinement level and a **`nested`** index with the formula `index_zuniq = (2 * index_nested + 1) * 4 ^ (29 - refinement_level)` <<K24>>.

The  **`refinement_level`** parameter must be omitted for the **`nuniq`** and **`zuniq`** schemes, for which the resolution of a cell is implied by the index alone.

A refinement level, either given by the  **`refinement_level`** parameter or implied by **`nuniq`** and **`zuniq`** indices, must be no greater than `29`, which is the highest refinement level for which all indices of any indexing scheme can be represented by 64-bit integers.

----

```markdown
Cells are indexed using either the **`nested`**, **`ring`**, **`nuniq`** or **`zuniq`** indexing scheme.

The **`ring`** scheme sorts pixels along iso-latitude rings from north to south; and **`nested`** indexing sorts pixels along the Z-order curve, resulting in spatial proximity for consecutively pixels.

For both the **`nested`** and **`ring`** schemes, the grid resolution is controlled by the refinement level parameter **`refinement_level`** <<GHB05>>.
The total number of cells on the sphere is `12 * (4 ^ refinement_level)`.
The lowest allowed refinement level is `0`, giving 12 cells globally.
The algorithm that assigns cell locations to **`nested`** or **`ring`** indices at a given refinement level is defined in <<GHB05>>. In this algorithm, a cell with index `0`, at any refinment level, always has its southern vertex on the meridian at 45 degrees east.

The **`nuniq`** and **`zuniq`** indexing schemes allow pixels from multiple refinement levels to be combined: **`nuniq`** effectively concatenates **`nested`** indices from lower to the higher refinement levels, resulting in spatial proximity for consecutive pixels within a refinement level but not across different refinement levels; and **`zuniq`** indexing sorts pixels of any refinement level along the Z-order curve, resulting in spatial proximity for consecutive pixels, regardless of their refinement level.

An index according to the **`nuniq`** indexing scheme can be computed from a refinement level and a **`nested`** index with the formula `index_nuniq = index_nested + 4 ^ (1 + refinement_level)` <<RH15>>.

An index according to the **`zuniq`** indexing scheme can be computed from a refinement level and a **`nested`** index with the formula `index_zuniq = (2 * index_nested + 1) * 4 ^ (29 - refinement_level)` <<K24>>.

The  **`refinement_level`** parameter must be omitted for the **`nuniq`** and **`zuniq`** schemes, for which the resolution of a cell is implied by the index alone.

A refinement level, either given by the  **`refinement_level`** parameter or implied by **`nuniq`** and **`zuniq`** indices, must be no greater than `29`, which is the highest refinement level for which all indices of any indexing scheme can be represented by 64-bit integers.
```

d70-t: Thanks for formulating this!
I've incorporated your proposed changes with slight modifications:

I've added "within each of the 12 global cells at refinement level `0`" to the description of the nested indexing Z-order scheme, and because of that addition swapped the ring-vs-nest paragraph with the refinement level & cell count paragraph.

appf.adoc
```diff
@@ -112,6 +112,47 @@ Note also that the values "x" and "y" are not case-sensitive.
...
 
+[[healpix]]
+=== HEALPix
+
+----
+grid_mapping_name = healpix
+----
+
+__Map parameters:__::
+* **`indexing_scheme`**
```
davidhassell:
```suggestion
* **`indexing_scheme`** - The HEALPix indexing scheme: One of **`nested`**, **`ring`**, **`nuniq`**, or **`zuniq`**
```

appf.adoc
```diff
@@ -112,6 +112,47 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+* **`refinement_level`** - Mandatory for `indexing_scheme` "nested" or "ring", must be omitted for "nuniq" or "zuniq".
```
davidhassell:
```suggestion
* **`refinement_level`** - The HEALPix refinement level: Mandatory for indexing scheme **`nested`** or **`ring`**, but must be omitted for indexing scheme **`nuniq`** or **`zuniq`**
```

appf.adoc
```diff
@@ -112,6 +112,47 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+The integer-valued HEALPix index coordinates are identified by the **`standard_name`** attribute value **`healpix_index`**.
+The HEALPix index coordinates may be stored in a coordinate, auxiliary, or scalar coordinate variable. Indexing schemes are optimized to allow applications efficient data access when the HEALPix indices are ordered strictly monotonically. Therefore it is recommended to store strictly monotonic HEALPix indices in a coordinate variable, which allows an application to assess their monotonicity without expensive checking.
++
+The coordinate system is defined by discrete cells on a sphere, arranged on a HEALPix grid.
+Each cell is represented by its cell index, which, together with the `indexing_scheme` and possible `refinement_level`, determines center coordinates (latitude and longitude) as well as boundaries.
+See <<GHB05>> for the calculation of center coordinates and cell boundary coordinates.
+
+__Notes:__::
+
+Cells are indexed using either the **`nested`**, **`ring`**, **`nuniq`** or **`zuniq`** indexing scheme.
```
davidhassell:
```suggestion
Cells are indexed using the scheme identified by the **`indexing_scheme`** parameter.
```

appf.adoc
```diff
@@ -454,6 +495,9 @@ The valid names are obtained by transforming the EPSG name using the following r
 This is to ensure that named datums can be correctly identified for precise datum transformations (see <<CF-WKT>> for more details).
 Corresponds to a OGC WKT DATUM node name.
 
+| **`indexing_scheme`** | S
+| The HEALPix cell indexing scheme, either **`nested`**, **`ring`**, **`nuniq`** or **`zuniq`**. See <<healpix>> for more detailed definitions.
```
davidhassell:
```suggestion
| The HEALPix cell indexing scheme, either **`nested`**, **`ring`**, **`nuniq`**, or **`zuniq`**. See <<healpix>> for more detailed definitions.
```

davidhassell (commented):
A few more things that occurred to me ... just on the clarity (?) of the text - no conceptual changes.

carloshorn:
Hi @tinaok,
This PR does not include authalic latitudes, but it will not exclude it.
The authalic latitude has a much wider scope and could be applicable beyond HEALPix. 
You could call it a week point of current CF conventions that by defining an ellipsoid it automatically implies geographic coordinates. And this is where you may start with a new issue that decouples the coordinate system from the earth model. It should allow you to use geodetic, geocentric, authalic, parametric, etc. coordinates for your selected grid mapping. Maybe the discussion within such an issue leads to the consensus that it is a DGGS specific problem, because classical projections do not need that or nobody wants this. Then it could be a special attribute of the HEALPix grid mapping, but it is too early to decide this. 
Once there is a way to inform about the chosen coordinates, this PR should be able to handle it.

davidhassell:
@d70-t  - the conformance changes look great, thanks.

tinaok:
@davidhassell, I agree that "body" is vague. I searched the document and didn’t find any prior use of that term. However, I do see use of expressions like "reference ellipsoid" or "ellipsoid (or sphere)."

We use one more time "sphere" (specifically when we mention the total number of cells on the sphere). I’d suggest updating this to "reference ellipsoid (or sphere)" instead of just "reference ellipsoid"

I think it’s useful to retain sphere here, because it aligns with the terminology in the original HEALPix paper. (Even though, realistically, we may never encounter a perfectly spherical system! ) By using "reference ellipsoid," rather than jubst "ellipsoid," I believe we also maintain a link with the grid mapping variable list you pointed out.

I also added 'surface ' to be precise.

appf.adoc
```diff
@@ -112,6 +112,47 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+See <<GHB05>> for the calculation of center coordinates and cell boundary coordinates.
+
+__Notes:__::
+
+Cells are indexed using the scheme identified by the **`indexing_scheme`** parameter.
++
+For both the **`nested`** and **`ring`** schemes, the grid resolution is controlled by the refinement level parameter **`refinement_level`** <<GHB05>>.
+The total number of cells on the surface of the sphere (or reference ellipsoid) is `12 * (4 ^ refinement_level)`.
+The lowest allowed refinement level is `0`, giving 12 cells globally.
+The algorithm that assigns cell locations to **`nested`** or **`ring`** indices at a given refinement level is defined in <<GHB05>>. In this algorithm, a cell with index `0`, at any refinment level, always has its southern vertex on the meridian at 45 degrees east.
```
carloshorn: Following: https://github.com/cf-convention/cf-conventions/issues/433#issuecomment-3517702797

I think we should insert the ellipsoid handling here. Maybe something like the following:

In the case of an ellipsoid, we construct the grid on the authalic sphere.
This means, that we extend the original algorithm to
```
lat_auth = geodetic_to_authalic(lat, a, b)
pix = lonlat_to_healpix(lon, lat_auth, level)
```
and 
```
lon, lat_auth = healpix_to_lonlat(pix, level)
lat = authalic_to_geodetic(lat_auth, a, b)
```
where `a` and `b` are semi-major and minor axis of the ellipsoid, ... `lonlat_to_healpix` and `healpix_to_lonlat` are the algorithms defined in `<<GHB05>>`

I think providing these pseudo code lines may help the reader. What do you think?

carloshorn: We should also mention that using authalic latitudes restores the equal area property. Otherwise, a (non-cartographer) reader could wonder why we did not use any other auxiliary latitude.

carloshorn: instead of `a` and `b` we may also just write `geodetic_to_authalic(lat, ellipse_params)`. Some people may think that `a` and `b` are required, then...

d70-t: If this helps (regarding `a` and `b`), there's already a [definition for `semi_major_axis` and `semi_minor_axis` in CF-conventions](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.12/cf-conventions.html#appendix-grid-mappings), and as far as I get the current state, these parameters would have to be specified anyways, if users want to communicate, that any grid mapping is using an ellipsoid.

So if we actually specify that we are always using authalic latitudes for HEALPix computations, I guess we can take these two parameters for granted (always, as it doesn't matter for a sphere).

carloshorn: Looking at the defined parameters, the `inverse_flattening` is sufficient for the transformation to authalic latitude, because auxiliary latitudes are scale independent. However, this detail is not too important. The point is, that a user should be aware of this transformation, because this will preserve the equal-area property on the ellipsoid, but it was/is our decision to generalize HEALPix this way on an ellipsoid, because it was originally defined on the sphere.

Nobody reacted until now, telling if they are using the almost equal area HEALPix variant, so if in doubt leave it out. I think a single choice aligns well with the goal of interoperability. Even though I invested time in understanding the almost equal area variant, I am fine with dropping it for this goal. My argument in the past was the non-existence of an analytic inverse, i.e. `authalic_to_geodetic`, which numerically would result in a root finding. After learning about the [series approximation using the third flattening](https://geographiclib.sourceforge.io/C++/doc/auxlat.html) (precise to floating point precision) and implementing it myself, I did not find a huge performance penalty.

If you agree, then let's think on how to incorporate this into the text without distraction by unnecessary details. Summarizing my thoughts about the important messages that we should transport: 
1. We decide here how we generalize HEALPix on an ellipsoid
2. We choose authalic latitudes to preserve the equal area property
3. We show explicitly where this transformation takes place having the spherical definition at hand, e.g. pseudo code.

Even though I think a hint to the existing series expansion would be beneficial, I guess it does not belong into the convention text. I am preparing material on HEALPix usage, where I will discuss it and also mention that the commonly used library *PROJ* also implements this choice. I should maybe also check other common projection libraries, e.g. *d3-geo*.

davidhassell: > 1.  We decide here how we generalize HEALPix on an ellipsoid

Agreed

> 2. We choose authalic latitudes to preserve the equal area property

Yes

> 3.    We show explicitly where this transformation takes place having the spherical definition at hand, e.g. pseudo code.

I actually find the pseudo code somewhat obfuscating - sorry! Function names themselves give little information, I feel.

Is it not enough to say that "_the grid cell locations on the surface of an ellipsoid are defined to give equal-area HEALPix cells when using authalic latitudes (which map the ellipsoid to a sphere with the same centre and  surface area as the ellipsoid)_".

carloshorn: No need to be sorry, as indicated by saying "e.g.", I am fine with any other way that states that this transformation is needed before using the defined HEALPix equations to transform geodetic coordinates to pixel index and back. Technically, this statement is enough. I just want the reader to not overlook this detail. I could also imagine that a clear point is made by giving the geodetic coordinates of the first tile at level 0 for WGS84. Something like: *"On WGS84, this results in the center of the first base tile being at geodetic latitude 41.938 degrees North, instead of 41.810 degrees North in the case of a sphere. The longitude remains at 45 degrees East in both cases."*

davidhassell: Thanks, @carloshorn - I like the note/example for WGS84 level 0, too. I'll suggest some text (I don't think we have any on this in the PR yet - I'll check first ...)

davidhassell:
```suggestion
The algorithm that assigns cell locations on a sphere to **`nested`** or **`ring`** indices at a given refinement level is defined in <<GHB05>>.
In this algorithm, a cell with index `0`, at any refinement level, always has its southernmost vertex on the meridian at 45 degrees east.
Cell locations on the surface of an ellipsoid are defined to give equal-area HEALPix cells when using authalic latitudes (which map locations on an ellipsoid to those on the sphere with the same surface area).
For instance, on the WGS84 ellipsoid this results in the center of the HEALPix cell at refinement level `0` with index `0` being at geodetic latitude 41.938 degrees north, instead of at 41.810 degrees north in the case of a sphere, whilst the longitude is at 45 degrees east in both cases.
```

d70-t: I'm still somewhat unconfortable mandating authalic latitudes (because I see that we'll run into people trying to compare healpix data on a sphere from models with healpix data on WGS84 from observations, and thos people might forget to re-grid from one healpix into the other healpix, which is necessary because the model was initialized using geodetic remapping, and thus implicitly uses geodetic healpix as an output).

But I totally see that authalic healpix is the right way to go if talking about observation data which truely is on WGS84 ellipsoids. So I guess authalic should be the default 🤷, and I'd go with your suggestions.

In any case, I've the feeling that we should try to formulate these sentences in a way which clearly states what our (the CF's) assumptions are, i.e. that users have no choice on the latitude conventions, but that CF conventions clearly define that authalic latitudes must be used. We might even want to say, that it's likely that these authalic latitudes shouldn't be presented to users, as likely everyone working with CF so far assumes every latitude is geodetic.

```suggestion
The algorithm that assigns cell locations on a sphere to **`nested`** or **`ring`** indices at a given refinement level is defined in <<GHB05>>.
In this algorithm, a cell with index `0`, at any refinement level, always has its southernmost vertex on the meridian at 45 degrees east.
If the **`healpix`** grid mapping is used on an ellipsoid, the latitudes returned by the <<GHB05>> algorithm must be interpreted as authalic latitude, instead of geodetic latitude (for a sphere, the two latitudes are the same).
This is necessary to retain the equal area property.
For instance, on the WGS84 ellipsoid this results in the center of the HEALPix cell at refinement level `0` with index `0` being at geodetic latitude 41.938 degrees north, instead of at 41.810 degrees north in the case of a sphere, whilst the longitude is at 45 degrees east in both cases.
```

davidhassell: That's great for me  - thanks!

appf.adoc
```diff
@@ -112,6 +112,47 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+* **`refinement_level`** - The HEALPix refinement level: Mandatory for indexing scheme **`nested`** or **`ring`**, but must be omitted for indexing scheme **`nuniq`** or **`zuniq`**
+
+__Map coordinates:__::
+The integer-valued HEALPix index coordinates are identified by the **`standard_name`** attribute value **`healpix_index`**.
+The HEALPix index coordinates may be stored in a coordinate, auxiliary, or scalar coordinate variable. Indexing schemes are optimized to allow applications efficient data access when the HEALPix indices are ordered strictly monotonically. Therefore it is recommended to store strictly monotonic HEALPix indices in a coordinate variable, which allows an application to assess their monotonicity without expensive checking.
++
+The coordinate system is defined by discrete cells on a reference ellipsoid (or sphere), arranged on a HEALPix grid.
+Each cell is represented by its cell index, which, together with the `indexing_scheme` and possible `refinement_level`, determines center coordinates (latitude and longitude) as well as boundaries.
+See <<GHB05>> for the calculation of center coordinates and cell boundary coordinates.
```
davidhassell, line 129: Should this paragraph go  _above_  the previous one?

history.adoc
```diff
@@ -12,6 +12,7 @@
 * {issues}583[Issue #583]: Correct "most rapidly varying dimension" in terminology section.
 * {issues}584[Issue #584]: Allow `Z` as time-zone offset, with a couple of examples, and allow time-zone offset with date alone, both being consistent with UDUNITS syntax.
 * {issues}508[Issue #508]: Introduce aggregation variables
+* {issues}433[Issue #433]: Convention for HEALPix grid parameters
```
JonathanGregory: The new entry is usually put at the top of the list, which is in reverse-historical order of being accepted. It doesn't matter that it's got a lower issue number.

conformance.adoc
```diff
@@ -372,10 +372,14 @@ If it is an auxiliary coordinate variable, it must be listed in the coordinates
...
+* The data types of the attributes of the grid mapping variable must be as specified in Table F.1 of Appendix F. +
 * If present, the **`crs_wkt`** attribute must be a text string conforming to the CRS WKT specification described in reference [OGC_CTS].
 * **`reference_ellipsoid_name`**, **`prime_meridian_name`**, **`horizontal_datum_name`** and **`geographic_crs_name`** must be all defined if any one is defined.
 * If **`projected_crs_name`** is defined then **`geographic_crs_name`** must be also.
+* If the value of the **`indexing_scheme`** attribute is **`nested`** or **`ring`**, then the **`refinement_level`** attribute must also be present.
+* If the value of the **`indexing_scheme`** attribute is **`nuniq`** or **`zuniq`**, then the **`refinement_level`** attribute must not be present.
+* The value of the **`refinement_level`** attribute must be an integer in the range 0-29.
+* A variable with standard name **`healpix_index`** must have an integer data type.
 
 *Recommendations:*
```
JonathanGregory, line 385: You could add a recommendation that an auxiliary coordinate variable with standard name `healpix_index`, of size greater than one, all of whose values are distinct, should be a coordinate variable instead (with values ordered monotonically).

appf.adoc
```diff
@@ -112,6 +112,51 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+* **`refinement_level`** - The HEALPix refinement level: Mandatory for indexing scheme **`nested`** or **`ring`**, but must be omitted for indexing scheme **`nuniq`** or **`zuniq`**
+
+__Map coordinates:__::
+The coordinate system is defined by discrete cells on a reference ellipsoid (or sphere), arranged on a HEALPix grid.
+Each cell is represented by its cell index, which, together with the `indexing_scheme` and possible `refinement_level`, determines center coordinates (latitude and longitude) as well as boundaries.
+See <<GHB05>> for the calculation of center coordinates and cell boundary coordinates.
++
+The integer-valued HEALPix index coordinates are identified by the **`standard_name`** attribute value **`healpix_index`**.
+The HEALPix index coordinates may be stored in a coordinate, auxiliary, or scalar coordinate variable. Indexing schemes are optimized to allow applications efficient data access when the HEALPix indices are ordered strictly monotonically. Therefore it is recommended to store strictly monotonic HEALPix indices in a coordinate variable, which allows an application to assess their monotonicity without expensive checking.
```
JonathanGregory: We usually take a new line for each sentence in `adoc`

appf.adoc
```diff
@@ -112,6 +112,51 @@ Note also that the values "x" and "y" are not case-sensitive.
...
+__Map parameters:__::
+* **`indexing_scheme`** - The HEALPix indexing scheme: One of **`nested`**, **`ring`**, **`nuniq`**, or **`zuniq`**
+* **`refinement_level`** - The HEALPix refinement level: Mandatory for indexing scheme **`nested`** or **`ring`**, but must be omitted for indexing scheme **`nuniq`** or **`zuniq`**
+
+__Map coordinates:__::
+The coordinate system is defined by discrete cells on a reference ellipsoid (or sphere), arranged on a HEALPix grid.
+Each cell is represented by its cell index, which, together with the `indexing_scheme` and possible `refinement_level`, determines center coordinates (latitude and longitude) as well as boundaries.
+See <<GHB05>> for the calculation of center coordinates and cell boundary coordinates.
++
+The integer-valued HEALPix index coordinates are identified by the **`standard_name`** attribute value **`healpix_index`**.
```
JonathanGregory: Are their values necessarily non-negative? If so, you could say it, and make it a conformance requirement as well.

appf.adoc
```diff
@@ -112,6 +112,51 @@ Note also that the values "x" and "y" are not case-sensitive.
...
++
+The **`ring`** scheme sorts pixels along iso-latitude rings from north to south; and **`nested`** indexing sorts pixels along the Z-order curve within each of the 12 global cells at refinement level `0`, resulting in spatial proximity for consecutive pixels.
++
+The **`nuniq`** and **`zuniq`** indexing schemes allow pixels from multiple refinement levels to be combined: **`nuniq`** effectively concatenates **`nested`** indices from lower to the higher refinement levels, resulting in spatial proximity for consecutive pixels within a refinement level but not across different refinement levels; and **`zuniq`** indexing sorts pixels of any refinement level along the Z-order curve, resulting in spatial proximity for consecutive pixels, regardless of their refinement level.
++
+An index according to the **`nuniq`** indexing scheme can be computed from a refinement level and a **`nested`** index with the formula `index_nuniq = index_nested + 4 ^ (1 + refinement_level)` <<RH15>>.
```
JonathanGregory: In this sentence and the next one, could you clarify that "**`nested`** index" and `index_nested` mean the cell index in a `nested` HEALpix grid. If this is so, it may seem obvious to you, but (as you can see) I am not completely certain on reading it.

conformance.adoc
davidhassell: I agree - we make this recommendation in the main text.

appf.adoc
d70-t: They are either non-negative, or it doesn't matter (because the number is interpreted as bit-pattern and the sign-bit isn't used). But in any case, the non-negative version seems to be the more accepted and elegant version. Thus I think we could make it a conformance requirement. I'll do so later, if no one complains.

conformance.adoc
d70-t: I've added the statement to the conformance document.

appf.adoc
d70-t: I tried to clarify the sentence.

appf.adoc
d70-t: I've updated the document.

conformance.adoc
davidhassell: I think the new addition need to go in "Recommendations:" section just below where it is now.

PR #609, opened 2025-09-30
Adding a dependabot yaml file to keep the workflows updated.

ChrisBarker-NOAA:
@sadielbartholomew: This was inspired by a commit you recently did to update an action.

I've added a dependabot file, which *should* automatically update the actions we are using for our publish workflow.

Nicer than having to go update those by hand.

I borrowed this from another project, so it *may* need some tweaking, but It should work as is.

When there is an update available, the dependabot creates a PR, which can then be merged.

NOTE: there don't seem to be any labels (or milestones) for this kind of infrastructure change .... maybe there should be?

@cf-convention/info-mgmt: pinging the team -- any one of you should be able to merge.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+# See https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/keeping-your-actions-up-to-date-with-dependabot
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


cofinoa:
Thanks @ChrisBarker-NOAA.

Looks good to me. This is a low-risk infrastructure improvement and I’m happy to see it merged unless there are objections.

I’ve added the `GitHub` and `enhancement` labels.

ChrisBarker-NOAA:
thanks!

PR #611, opened 2025-11-10
Revision to leap seconds and calendar conventions (v2)

JonathanGregory:
**This PR is a work in progress and not yet ready for general comments**

Closed in favour of #612.

### DIFF
--- a/ch04.adoc
+++ b/ch04.adoc
@@ -361,7 +361,8 @@ A non-zero time zone offset is not allowed in this calendar.
 **`tai`**:: A Gregorian calendar __without__ leap seconds that is based on International Atomic Time (TAI), which began on 1958-01-01 0:0:0.
 Earlier datetimes than this are not allowed in the **`tai`** calendar.
 For any given instant, the **`tai`** datetime is ahead of the **`utc`** datetime, where "ahead" means the same as it does when describing a timezone to the east as being ahead of one to the west.
-The difference between the two datetimes (**`tai`** minus **`utc`**) for a given instant of time is 10 seconds plus the net number of leap seconds introduced since 1972-01-01.
+The difference between the two datetimes (**`tai`** minus **`utc`**) for a given instant of time is 10 seconds plus the net number of leap seconds introduced since the instant with the datetime 1972-01-10 in the **`tai`** calendar, when the current definition of UTC came into force.
+This instant has the datetime 1972-01-01 00:00:00 in the **`utc`** calendar.
 Moreover, a given datetime in the **`tai`** calendar represents an instant that is earlier than the same datetime in the **`utc`** calendar.
 A non-zero time zone offset is not allowed in this calendar.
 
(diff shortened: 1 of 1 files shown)


ch04.adoc
```diff
@@ -204,310 +220,236 @@ Optionally, a time coordinate variable may be indicated additionally by providin
...
-**`tai`**:: A Gregorian calendar __without__ leap seconds that is based on International Atomic Time (TAI).
-Datetimes before 1958-01-01 0:0:0 are not allowed in this calendar.
+**`tai`**:: A Gregorian calendar __without__ leap seconds that is based on International Atomic Time (TAI), which began on 1958-01-01 0:0:0.
+Earlier datetimes than this are not allowed in the **`tai`** calendar.
 For any given instant, the **`tai`** datetime is ahead of the **`utc`** datetime, where "ahead" means the same as it does when describing a timezone to the east as being ahead of one to the west.
-The difference between the two datetimes for a given instant of time is the net number of leap seconds introduced since 1958-01-01.
-The difference was zero on that instant, when both calendars began.
-This means that a given datetime in the **`tai`** calendar represents an instant that is later than the same datetime in the **`utc`** calendar.
-See also <<leap-seconds>>.
+Moreover, a given datetime in the **`tai`** calendar represents an instant that is later than the same datetime in the **`utc`** calendar.
```
pvanlaake:
```suggestion
Moreover, a given datetime in the **`tai`** calendar represents an instant that is earlier than the same datetime in the **`utc`** calendar.
```

JonathanGregory: Thanks for noticing this. I think you're right, and therefore the corresponding statement for UTC is also wrong. It should be, "Moreover, a given datetime in the `utc` calendar represents an instant that is later than the same datetime in the `tai` calendar." The fact that I got it wrong suggests to me that it's worth spelling it out, provided we do so correctly! I'll change them both. Also it now seems to make more sense to swap the order of two of the sentences, thus (for TAI):

> For any given instant, the **`tai`** datetime is ahead of the **`utc`** datetime, where "ahead" means the same as it does when describing a timezone to the east as being ahead of one to the west. The difference between the two datetimes (**`tai`** minus **`utc`**) for a given instant of time is 10 seconds plus the net number of leap seconds introduced since 1972-01-01. Moreover, a given datetime in the **`tai`** calendar represents an instant that is earlier than the same datetime in the **`utc`** calendar.

ChrisBarker-NOAA: Honestly, I think we're putting more that necessary description in the CF doc -- if you don't know what TAI and UTC mean, you shouldn't be using them :-(.  -- but anyway, I think an example is in order -- examples are always more clear to users. Something like:

For example: `2025-11-13T21:00:43 TAI` is the same instant as: `2025-11-13T21:00:10 UTC`, as there have been 33 leap seconds introduced as of that date.

[Example taken from: http://www.csgnetwork.com/timetaidispcalc.html, which I presume is correct]

JonathanGregory: I think you're commenting on the `tai` entry. Rather than adding an extra example, we can mirror what is said for `utc`, as follows (new and changed text in italics here in order to highlight it, but not in the document):

> For any given instant, the **`tai`** datetime is ahead of the **`utc`** datetime, where "ahead" means the same as it does when describing a timezone to the east as being ahead of one to the west. The difference between the two datetimes (**`tai`** minus **`utc`**) for a given instant of time is 10 seconds plus the net number of leap seconds introduced since _the instant with the datetime 1972-01-10 in the **`tai`** calendar, when the current definition of UTC came into force. This instant has the datetime 1972-01-01 00:00:00 in the **`utc`** calendar._ Moreover, a given datetime in the **`tai`** calendar represents an instant that is earlier than the same datetime in the **`utc`** calendar.

Thanks for the comment. I agree that if you're going to _write_ data in UTC or TAI you should already know what they mean. However, we can't depend on _users_ of the data being familiar with the detailed definitions. Yet they need to know them in order to interpret the data correctly.

JonathanGregory:
I've just corrected a mistake I made when adding Lars's table to the document. In this version, he had "yes" for "Leap days" in the non-model calendars. In order to give more information, I changed "yes" to "Julian" or "Gregorian",  but I forgot that standard is both, so I've now put "Julian/Gregorian" for its entry.

ch04.adoc
ChrisBarker-NOAA: This looks OK to me.

> Thanks for the comment. I agree that if you're going to write data in UTC or TAI you should already know what they mean. However, we can't depend on users of the data being familiar with the detailed definitions. Yet they need to know them in order to interpret the data correctly.

Or have tools that know :-) 

Anyway, I generally think more info is better than less.

PR #630, opened 2026-01-28
Minor fixes for Zenodo release of CF-1.13

davidhassell:
No associated issue.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -134,9 +134,11 @@ authors:
 - affiliation: NCAS and STFC
   family-names: Massey
   given-names: Neil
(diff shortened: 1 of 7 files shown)


cofinoa:
Thanks @davidhassell 

I’ve reviewed this PR and it looks good to me. These are the final editorial tweaks we discussed for the CF-1.13 Zenodo artifacts (name/affiliation, ORCID propagation, and the minor revision-history fix), and they align with the PDFs you sent.

Unless anyone spots an issue, I think it is prudent to merge this into `main` so the repo reflects what we are publishing.

Happy for you to merge, or I can do it if you prefer.

PR #637, opened 2026-03-09
Revision of section 4.3 to improve its clarity

### LINKED ISSUE #635: revision of 4.3 to improve clarity
In [Discussion 462](https://github.com/orgs/cf-convention/discussions/462) @ilaflott asked about a possible inconsistency in 4.3. @taylor13 explained how the text is to be understood, and that it's not self-contradictory, but commented that it is poorly organised. I agree that it's not well-organised and therefore could be confusing; also, it's rather repetitive. I therefore propose to reword it as follows, where ~~strikethrough~~ means deletion and **bold** means new text, although some of the new text is existing text rearranged.

Is this version clear?

### 4.3 Vertical (Height or Depth) Coordinate

Variables representing ~~dimensional height or depth axes must always explicitly include the **`units`** attribute; there is no default value. A vertical coordinate will be identifiable by:~~ **vertical coordinates must have either**

* ~~the presence of~~ **a `positive`** attribute with a value of **`up`** or **`down`** (case insensitive), which respectively mean that increasingly positive coordinate values indicate higher locations i.e. further from the centre of the Earth (**`positive="up"`**) or lower locations (**`positive="down"`**); or
* **a `units` attribute containing units of pressure (as determined by UDUNITS).** If the **`units`** attribute value is a valid pressure unit the default value of the **`positive`** attribute is **`down`**.

~~The direction of positive (i.e., the direction in which the coordinate values are increasing), whether up or down, cannot in all cases be inferred from the units.
The direction of positive is useful for applications displaying the data.
For this reason the attribute **`positive`** as defined in the COARDS conventions is required if the vertical axis units are not a valid unit of pressure (as determined by the UDUNITS package <<UDUNITS>>) -- otherwise its inclusion is optional.
The **`positive`** attribute may have the value **`up`** or **`down`** (case insensitive).~~

The **`positive`** attribute may be applied **both** to ~~either~~ coordinate variables ~~or~~ **and to** auxiliary coordinate variables that contain vertical coordinate data.

For example, if an oceanographic netCDF file encodes the depth of the surface as 0 and the depth of 1000 meters as 1000 then the axis would use attributes as follows: 

```
axis_name:units = "meters" ; 
axis_name:positive = "down" ; 	
```

If, on the other hand, the depth of 1000 meters were represented as -1000 then the value of the **`positive`** attribute would have been **`up`**.

Optionally, the vertical type may be indicated additionally by providing the **`standard_name`** attribute with an appropriate value, and/or the **`axis`** attribute with the value **`Z`**. 
If both **`positive`** and **`standard_name`** are provided, it is recommended that they should be consistent. 
For instance, if a depth of 1000 metres is represented by -1000 and **`positive`** is **`up`**, it would be inconsistent to give the **`standard_name`** as **`depth`**, whose definition (vertical distance below the surface) implies positive down. 
If an application detects such an inconsistency, the user should be warned, and the **`positive`** attribute should be used to determine the sign convention.

#### 4.3.1 Dimensional Vertical Coordinate

Variables representing dimensional vertical coordinates **(i.e. not dimensionless numbers, but having a unit of measure)** ~~for depth or height~~ must always explicitly include the **`units`** attribute; there is no default value.
The ~~acceptable~~ **`units`** for a vertical ~~(depth or height)~~ coordinate variable must **be valid according to UDUNITS** ~~representation of one of~~; acceptable possibilities include the following:

* units of pressure;
~~For vertical axes~~ the most commonly used of these include **`bar`**, **`millibar`**, **`decibar`**, **`atmosphere (atm)`**, **`pascal (Pa)`**, and **`hPa`**.
* units of length;
~~For vertical axes~~ the most commonly used of these include **`meter (metre, m)`**, and **`kilometer (km)`**.
* other units that may under certain circumstances reference vertical position such as units of density or temperature.

~~Plural forms are also acceptable.~~ _@JonathanGregory: True, but it's always true, since UDUNITS allows it, so it seems odd to say it here._

##### 4.3.2 Dimensionless Vertical Coordinate

The **`units`** attribute is not required for dimensionless coordinates.
For backwards compatibility with COARDS the **`units`** attribute **for dimensionless vertical coordinates** may take one of the values: **`level`**, **`layer`**, or **`sigma_level`**.
These values are not recognized by the UDUNITS package, and are considered a deprecated feature in the CF conventions.

JonathanGregory:
See issue #635 for discussion of these changes.

# Release checklist

### DIFF
--- a/ch04.adoc
+++ b/ch04.adoc
@@ -90,13 +90,13 @@ Coordinates of longitude with respect to a rotated pole should be given units of
 [[vertical-coordinate, Section 4.3, "Vertical Coordinate"]]
 === Vertical (Height or Depth) Coordinate
 
-Variables representing dimensional height or depth axes must always explicitly include the **`units`** attribute; there is no default value.
+Variables representing vertical coordinates must have either
 
-The direction of positive (i.e., the direction in which the coordinate values are increasing), whether up or down, cannot in all cases be inferred from the units.
-The direction of positive is useful for applications displaying the data.
-For this reason the attribute **`positive`** as defined in the COARDS conventions is required if the vertical axis units are not a valid unit of pressure (as determined by the UDUNITS package <<UDUNITS>>) -- otherwise its inclusion is optional.
-The **`positive`** attribute may have the value **`up`** or **`down`** (case insensitive).
-This attribute may be applied to either coordinate variables or auxiliary coordinate variables that contain vertical coordinate data.
+* a **`positive`** attribute with a value of **`up`** or **`down`** (case insensitive), which respectively mean that increasingly positive coordinate values indicate higher locations i.e. further from the centre of the Earth (**`positive="up"`**) or lower locations (**`positive="down"`**); or
+
+* a **`units`** attribute containing units of pressure (as determined by UDUNITS). If the **`units`** attribute value is a valid pressure unit the default value of the **`positive`** attribute is **`down`**.
+
+The **`positive`** attribute may be applied both to coordinate variables and to auxiliary coordinate variables that contain vertical coordinate data.
 
 For example, if an oceanographic netCDF file encodes the depth of the surface as 0 and the depth of 1000 meters as 1000 then the axis would use attributes as follows: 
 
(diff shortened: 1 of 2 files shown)


ch04.adoc
```diff
@@ -90,13 +90,13 @@ Coordinates of longitude with respect to a rotated pole should be given units of
...
+Variables representing vertical coordinates must have either
 
-The direction of positive (i.e., the direction in which the coordinate values are increasing), whether up or down, cannot in all cases be inferred from the units.
-The direction of positive is useful for applications displaying the data.
-For this reason the attribute **`positive`** as defined in the COARDS conventions is required if the vertical axis units are not a valid unit of pressure (as determined by the UDUNITS package <<UDUNITS>>) -- otherwise its inclusion is optional.
-The **`positive`** attribute may have the value **`up`** or **`down`** (case insensitive).
-This attribute may be applied to either coordinate variables or auxiliary coordinate variables that contain vertical coordinate data.
+* a **`positive`** attribute with a value of **`up`** or **`down`** (case insensitive), which respectively mean that increasingly positive coordinate values indicate higher locations i.e. further from the centre of the Earth (**`positive="up"`**) or lower locations (**`positive="down"`**); or
```
larsbarring:
```suggestion
the attribute **`positive`** that takes either the value **`up`** or **`down`** (case insensitive), which respectively mean that increasingly positive coordinate 
```

